PR #2, opened 2021-03-19
added DB naming requirements

Max-Pochepets:


README.md
```diff
@@ -20,6 +20,7 @@ public class Main {
     }
 }
 ```
+- Your table in DB should be called according to model name, in plural and lower case. For table's columns: `modelName_columnName`.
```
RomanDubovskyi, line 23: we don't use camelCase in naming. Btw if we want to transfer this point from checklist to task description I would add bad/good example as well - they are easy to understand

RomanDubovskyi (changes requested):

README.md
boroda4436: +1 to bad/good example

Max-Pochepets:
Don't think it's a good idea to show good/bad examples in task description. We have checklist for this.

RomanDubovskyi (approved):

PR #11, opened 2021-05-12
initial commit

fotomassimo:


src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,131 @@
...
+        } catch (SQLException exception) {
+            throw new DataProcessingException("Can't create manufacturer", exception);
+        }
+    }
+
+    private boolean checkForNullId(Long id) {
+        if (id == null) {
+            throw new DataProcessingException("Manufacturer id can't be null for DB operation");
+        }
+        return true;
```
LayRipper, line 129: I believe this check is sort of excess as you will not get autogenerated id as a null unless it is set purposely

src/main/java/mate/jdbc/exception/DataProcessingException.java
```diff
@@ -0,0 +1,11 @@
+package mate.jdbc.exception;
+
+public class DataProcessingException extends RuntimeException {
+    public DataProcessingException(String message) {
+        super(message);
+    }
```
LayRipper, line 6: one constructor is enough for the purpose of this task

src/main/java/mate/jdbc/model/Manufacturer.java
```diff
@@ -0,0 +1,46 @@
...
+public class Manufacturer {
+    private Long id;
+    private String name;
+    private String country;
+
+    public Manufacturer(Long id, String name, String country) {
+        this.id = id;
+        this.name = name;
+        this.country = country;
+    }
```
LayRipper, line 12: do we need this id in the constructor? it causes only confusion and is not used anywhere

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,131 @@
...
+
+    @Override
+    public boolean delete(Long id) {
+        checkForNullId(id);
+        String deleteQuery = "UPDATE manufacturers SET is_deleted = true WHERE id = ?;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement deleteManufacturerStatement = connection
+                        .prepareStatement(deleteQuery)) {
+            deleteManufacturerStatement.setLong(1, id);
+            return deleteManufacturerStatement.executeUpdate() >= 1;
```
LayRipper, line 108: will this value ever be > 1 ?

LayRipper (changes requested):

LayRipper (approved):

src/main/java/mate/jdbc/dao/BasicDao.java
```diff
@@ -0,0 +1,16 @@
+package mate.jdbc.dao;
+
+import java.util.List;
+import java.util.Optional;
+
+public interface BasicDao<T, K> {
```
KseniiaMakarova, line 6:
```suggestion
public interface AbstractDao<T, K> {
```

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,125 @@
...
+            Manufacturer manufacturer = new Manufacturer(resultSet.getString("name"),
+                    resultSet.getString("country"));
+            manufacturer.setId(resultSet.getObject("id", Long.class));
+            return manufacturer;
+        } catch (SQLException exception) {
+            throw new DataProcessingException("Can't create manufacturer", exception);
+        }
+    }
+
+    private boolean checkForNullId(Long id) {
```
KseniiaMakarova, line 119:
```suggestion
    private void checkIdForNull(Long id) {
```

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,125 @@
...
+            return manufacturer;
+        } catch (SQLException exception) {
+            throw new DataProcessingException("Can't create manufacturer "
+                    + manufacturer.getName() + " in DB", exception);
+        }
+    }
+
+    @Override
+    public Optional<Manufacturer> get(Long id) {
+        checkForNullId(id);
```
KseniiaMakarova, line 41: have you tried making a query in MySQL Workbench where you select by NULL id?
if this query is not possible, then there is no real need in this private method

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    @Override
+    public Manufacturer update(Manufacturer manufacturer) {
+        checkForNullId(manufacturer.getId());
+        String updateQuery = "UPDATE manufacturers SET name = ?, country = ? WHERE id = ? "
+                + "AND is_deleted = FALSE;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement updateManufacturerStatement = connection
+                        .prepareStatement(updateQuery, Statement.RETURN_GENERATED_KEYS)) {
```
KseniiaMakarova, line 81: do we need generated keys here?

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,125 @@
...
+
+    @Override
+    public boolean delete(Long id) {
+        checkForNullId(id);
+        String deleteQuery = "UPDATE manufacturers SET is_deleted = true WHERE id = ?;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement deleteManufacturerStatement = connection
+                        .prepareStatement(deleteQuery)) {
+            deleteManufacturerStatement.setLong(1, id);
+            return deleteManufacturerStatement.executeUpdate() == 1;
```
KseniiaMakarova, line 101:
```suggestion
            return deleteManufacturerStatement.executeUpdate() > 0;
```

src/main/java/mate/jdbc/model/Manufacturer.java
```diff
@@ -0,0 +1,45 @@
+package mate.jdbc.model;
+
+public class Manufacturer {
+    private Long id;
+    private String name;
+    private String country;
+
+    public Manufacturer(String name, String country) {
```
KseniiaMakarova, line 8: let's create an empty constructor as well

src/main/java/mate/jdbc/util/ConnectionUtil.java
```diff
@@ -0,0 +1,31 @@
...
+    private static final Properties properties = new Properties();
+    private static final String PROPERTIES_FILE_PATH = "src/main/resources/MySQL_DB.properties";
+
+    static {
+        try {
+            properties.load(new FileReader(PROPERTIES_FILE_PATH));
+            Class.forName((String) properties.get("db.driver"));
+        } catch (ClassNotFoundException | IOException exception) {
+            throw new RuntimeException("Can't load JDBC driver for MySQL", exception);
+        }
```
KseniiaMakarova, line 20: let's better have 2 catch blocks - the reasons of exceptions are different

KseniiaMakarova (changes requested):
great job! added minor comments

src/main/java/mate/jdbc/dao/BasicDao.java
fotomassimo: done

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
fotomassimo: done

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
fotomassimo: I tried. Before:  https://prnt.sc/12vft2o
After:  https://prnt.sc/12vfui6
Seems like working. The reason I put this check that I get NPE if we call RUD methods with input id = null before putting entity into DB. So I replacing NPE with more informative exception. But I understand id = null does not occur in this task, so I better remove this check

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
fotomassimo: done

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
fotomassimo: done

src/main/java/mate/jdbc/model/Manufacturer.java
fotomassimo: done

src/main/java/mate/jdbc/util/ConnectionUtil.java
fotomassimo: done

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
KseniiaMakarova: great, thank you!

KseniiaMakarova (approved):

PR #80, opened 2021-08-09
Implemented JDBC

DanaKhromenko:
- established connection to Database;
- created init_db.sql file;
- created Manufacturer model;
- created DAO layer for Manufacturer model;
- used @Dao annotation for Dao implementation;
- called CRUD methods in the main method.

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -1,7 +1,32 @@
 package mate.jdbc;
 
+import mate.jdbc.dao.ManufacturerDao;
+import mate.jdbc.dao.ManufacturerDaoImpl;
+import mate.jdbc.model.Manufacturer;
+
 public class Main {
     public static void main(String[] args) {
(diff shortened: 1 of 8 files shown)


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,32 @@
 package mate.jdbc;
 
+import mate.jdbc.dao.ManufacturerDao;
+import mate.jdbc.dao.ManufacturerDaoImpl;
+import mate.jdbc.model.Manufacturer;
+
 public class Main {
     public static void main(String[] args) {
+        ManufacturerDao manufacturerDao = new ManufacturerDaoImpl();
```
RomanDubovskyi, line 9: тут можна інжектор використати 
```suggestion
         ManufacturerDao manufacturerDao =
                (ManufacturerDao) injector.getInstance(ManufacturerDao.class);
```

RomanDubovskyi (changes requested):
Great job, use injector instead creating instances, rest LGTM

PR #89, opened 2021-09-21
solution task "jv-JDBC"

Chupiak:
I think method "update" will be accept another arguments. 
Method will be changed to next view : 

Manufacturer update(long id, String name, String country);

### DIFF
--- a/src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
+++ b/src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
@@ -1,0 +1,119 @@
+package mate.jdbc.dao;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class ManufacturerDaoImpl implements ManufacturerDao {
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        String insertRequest = "INSERT INTO manufacturers(name,country) values(?,?);";
+
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement createStatement = connection.prepareStatement(insertRequest,
+                        Statement.RETURN_GENERATED_KEYS)) {
+
+            createStatement.setString(1,manufacturer.getName());
+            createStatement.setString(2,manufacturer.getCountry());
+            createStatement.executeUpdate();
+            ResultSet generatedKeys = createStatement.getGeneratedKeys();
+            if (generatedKeys.next()) {
+                Long id = generatedKeys.getObject(1,Long.class);
+                manufacturer.setId(id);
+            }
+
+        } catch (SQLException e) {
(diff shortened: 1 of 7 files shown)


src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,119 @@
...
+                        Statement.RETURN_GENERATED_KEYS)) {
+            createFormatStatement.setObject(1,id);
+            return createFormatStatement.executeUpdate() >= 1;
+        } catch (SQLException e) {
+            throw new RuntimeException("Can't delete from DB " + id,e);
+        }
+    }
+
+    @Override
+    public Manufacturer update(Manufacturer manufacturer) {
```
RomanDubovskyi, line 54: basically we can make this method void but usually void methods are avoided because they are hard to test, anyway it's not crucial

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,119 @@
...
+                    "SELECT * FROM manufacturers WHERE is_deleted = 0");
+            while (resultSet.next()) {
+                Long id = resultSet.getLong("id");
+                String name = resultSet.getString("name");
+                String country = resultSet.getString("country");
+
+                Manufacturer manufacturer = new Manufacturer();
+                manufacturer.setId(id);
+                manufacturer.setName(name);
+                manufacturer.setCountry(country);
```
RomanDubovskyi, line 107: this code is duplicated, let's move forming object from ResultSet into private method

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,119 @@
...
+
+                allManufacturer.add(manufacturer);
+                System.out.println(manufacturer);
```
RomanDubovskyi, line 110: we don't need sout here

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,119 @@
...
+
+            }
+        } catch (SQLException e) {
+            throw new RuntimeException("Can't get all rows from DB",e);
```
RomanDubovskyi, line 114:
```suggestion
            throw new RuntimeException("Can't get all Manufacturers from DB",e);
```

RomanDubovskyi (changes requested):

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,118 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement createStatement = connection.prepareStatement(insertRequest,
+                        Statement.RETURN_GENERATED_KEYS)) {
+
+            createStatement.setString(1,manufacturer.getName());
+            createStatement.setString(2,manufacturer.getCountry());
+            createStatement.executeUpdate();
+            ResultSet generatedKeys = createStatement.getGeneratedKeys();
+            if (generatedKeys.next()) {
+                Long id = generatedKeys.getObject(1,Long.class);
```
RomanDubovskyi, line 30:
```suggestion
                Long id = generatedKeys.getObject(1, Long.class);
```

RomanDubovskyi (approved):

PR #104, opened 2021-09-30
jdbc HW-1

dogovor:


src/main/java/mate/jdbc/connectionutil/ConnectionUtil.java
```diff
@@ -0,0 +1,29 @@
+package mate.jdbc.connectionutil;
+
```
Zbarazkyy, line 2: This pact is better named simply util, perhaps you will have more classes that can be put here

src/main/java/mate/jdbc/model/Manufacturer.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return "Manufacturer{" + "id="
+                    + id + ", name='"
+                    + name + '\'' + ", country='"
+                    + country + '\'' + '}';
+    }
+}
```
Zbarazkyy, line 68: there is no package named resources where there will be a file with an sql query to create a database

Zbarazkyy (commented):
two remarks 
1) About the sql query file
2) rename the package, for more extensive use

Zbarazkyy (approved):

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,28 @@
...
+        Manufacturer bucketWithBolts = new Manufacturer("ZAZ", "Ukraine");
+        Manufacturer luxury = new Manufacturer("Maserati", "Italy");
+        Manufacturer solidity = new Manufacturer("Toyota Motor Corporation", "Japan");
+        Manufacturer janky = new Manufacturer("GM", "USA");
+        manufacturerDao.create(bucketWithBolts);
+        manufacturerDao.create(luxury);
+        manufacturerDao.create(solidity);
+        manufacturerDao.create(janky);
+        System.out.println(manufacturerDao.getAll().toString());
+        manufacturerDao.delete(4L);
```
MaliukDaria, line 22:
```suggestion
        manufacturerDao.delete(manufacturerDao.getId());
```
It's better not to write entity `id's` explicitly

MaliukDaria (approved):

PR #107, opened 2021-09-30
Yelinskyi

Yelinskyi:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,42 @@
...
+        manufacturer = new Manufacturer("Rolls Royce", "UK");
+        manufacturerDao.create(manufacturer);
+        // Show all items in the table DB:
+        allManufacturers = manufacturerDao.getAll();
+        allManufacturers.forEach(System.out::println);
+        // Show all items in the table DB:
+        allManufacturers = manufacturerDao.getAll();
+        allManufacturers.forEach(System.out::println);
+        // Is deleted instance in DB:
+        /*manufacturerDao.delete(15L);*/
```
Max-Pochepets, line 31: why is this commented?

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,144 @@
...
+                PreparedStatement getInstanceManufacturersStatement = connection.prepareStatement(
+                        getManufacturerRequest)) {
+            getInstanceManufacturersStatement.setLong(1, id);
+            ResultSet resultSet = getInstanceManufacturersStatement.executeQuery();
+            if (resultSet.next()) {
+                manufacturer = createManufacturerFromDataDB(resultSet);
+            } else {
+                throw new CustomJdbcException(("The manufacturer with id = "
+                        + id + " doesn't exists!"));
+            }
```
Max-Pochepets, line 61: no need to throw this exception

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,144 @@
...
+            return Optional.ofNullable(manufacturer);
+        } catch (SQLException e) {
+            throw new CustomJdbcException("Can't get an instance from DB with id = " + id, e);
```
Max-Pochepets, line 64:
```suggestion
            throw new DataProcessingException("Can't get an instance from DB with id = " + id, e);
```
image

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,144 @@
...
+        } catch (NoSuchElementException e) {
+            throw new CustomJdbcException("The instance with id = "
+                    + manufacturer.getId() + " doesn't exists!", e);
+        }
+        Optional<Manufacturer> optional = get(manufacturer.getId());
+        if (optional.isPresent()) {
+            return get(manufacturer.getId()).get();
+        }
+        throw new CustomJdbcException(("The manufacturer with id = "
+                + manufacturer.getId() + " doesn't exists!"));
```
Max-Pochepets, line 114: do we need all this?

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,144 @@
...
+    @Override
+    public boolean delete(Long id) {
+        String deleteManufacturerRequest =
+                "UPDATE manufacturers SET is_deleted = true WHERE id = ?";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement createManufacturerStatement =
+                        connection.prepareStatement(deleteManufacturerRequest,
+                        Statement.RETURN_GENERATED_KEYS)) {
+            createManufacturerStatement.setLong(1, id);
+            return createManufacturerStatement.executeUpdate() != 0;
```
Max-Pochepets, line 126: can here be a negative value?

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,144 @@
...
+    @Override
+    public Manufacturer update(Manufacturer manufacturer) {
+        String updateManufacturerRequest =
+                "UPDATE manufacturers "
+                        + "SET name = ?, country = ? "
+                        + "WHERE id = ? AND is_deleted = false";
+        try {
+            Connection connection = ConnectionUtil.getConnection();
+            PreparedStatement updateManufacturerStatement =
+                    connection.prepareStatement(updateManufacturerRequest);
```
Max-Pochepets, line 96: why not use try-with-resources?

Max-Pochepets (changes requested):

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
Yelinskyi: I think 'Yes'.
I can return 'manufacturer' but I wouldn't know if the table changes.

But if you have different opinion I'll do it.

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
Yelinskyi: I fogot to do it.

PR #131, opened 2021-10-04
Created a method to work with the database manufacturer

ivashchenko5:


src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,122 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getAllManufacturersStatement =
+                        connection.prepareStatement(getAllManufactureRequest)) {
+            ResultSet resultSet = getAllManufacturersStatement.executeQuery();
+            while (resultSet.next()) {
+                String name = resultSet.getString(NAME_COLUMN);
+                String country = resultSet.getString(COUNTRY_COLUMN);
+                Long id = resultSet.getObject(ID_COLUMN, Long.class);
+                Manufacturer manufacturer = new Manufacturer(name, country);
+                manufacturer.setId(id);
```
Sinelife, line 36: You has similar code in getAll methods. 
So it would be better to create method which will get manufacturer from ResultSet.

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,28 @@
...
+        Manufacturer firstManufacturer = new Manufacturer();
+        firstManufacturer.setName("Vito");
+        firstManufacturer.setCountry("USA");
+        System.out.println(manufacturerDao.create(firstManufacturer));
+        Manufacturer secondManufacturer = new Manufacturer();
+        secondManufacturer.setName("Bogdan");
+        secondManufacturer.setCountry("Ukraine");
+        manufacturerDao.create(secondManufacturer);
+        System.out.println(manufacturerDao.get(5L));
+        manufacturerDao.delete(4L);
```
Sinelife, line 22: for testing method delete it will be better to not just put id as a number.
It will be better to chose specific manufacturer and put his id. For example
manufacturerDao.delete(secondManufacturer.getId());

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,122 @@
...
+    public Optional<Manufacturer> get(Long id) {
+        String getManufacturerRequest
+                = "SELECT * FROM manufacturers WHERE is_deleted = false AND id = ?";
+        Manufacturer manufacturer = null;
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getManufacturerStatement =
+                            connection.prepareStatement(getManufacturerRequest)) {
+            getManufacturerStatement.setLong(1, id);
+            ResultSet resultSet = getManufacturerStatement.executeQuery();
+            while (resultSet.next()) {
```
Sinelife, line 55: you get here only one manufacturer so you can change while to if.

Sinelife (changes requested):

Sinelife (approved):

Resci (approved):

PR #177, opened 2021-12-09
Wrote an assignment jv-jdbc.

ASelivonchyk:
// injector not working, is it normal?
- created Manufacturer class as manufacturer model;
- created ManufacturerDao with CRUD methods and implement it in class ManufacturerDaoImpl;
- created ConnectionUtil class which establishes connection to DB;
- created custom exception - DataProcessingException;
- made init_db.sql with parameters to create DB and table;
- in Main class call all of ManufacturerDao methods;

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -8,8 +8,8 @@ public class Main {
     private static final Injector injector = Injector.getInstance("mate.jdbc");
 
     public static void main(String[] args) {
-        ManufacturerDao manufacturerDao =
-                (ManufacturerDao) injector.getInstance(ManufacturerDao.class);
+        ManufacturerDao manufacturerDao = (ManufacturerDao)
+                injector.getInstance(ManufacturerDao.class);
         Manufacturer volkswagen = manufacturerDao
                 .create(new Manufacturer("Volkswagen", "Germany"));
         Manufacturer skoda = manufacturerDao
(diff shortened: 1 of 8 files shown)


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,25 @@
...
+        Manufacturer volkswagen = manufacturerDao
+                .create(new Manufacturer("Volkswagen", "Germany"));
+        Manufacturer skoda = manufacturerDao
+                .create(new Manufacturer("Skoda", "Czech Republic"));
+        Manufacturer hyundai = manufacturerDao
+                .create(new Manufacturer("Hyundai", "Republic of Korea"));
+        manufacturerDao.getAll().forEach(System.out::println);
+        hyundai.setCountry("Korea");
+        manufacturerDao.update(hyundai);
+        System.out.println(manufacturerDao.get(3L));
```
MaliukDaria, line 22:
```suggestion
        System.out.println(hundai.getId());
```

Better not to get attached to a certain id, we can get it from Manufacturer entity

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,25 @@
...
+        System.out.println(manufacturerDao.delete(1L));
```
MaliukDaria, line 23: same here

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,106 @@
...
+
+@Dao
+public class ManufacturerDaoImpl implements ManufacturerDao {
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        String insertRequest = "INSERT INTO manufacturers(name, country) values(?, ?);";
+        try (var connection = ConnectionUtil.getConnection();
+                var createManufacturerStatement = connection
+                        .prepareStatement(insertRequest, Statement.RETURN_GENERATED_KEYS)) {
```
MaliukDaria, line 23: I think we should not use `var` here(same for other methods)

MaliukDaria (changes requested):

MaliukDaria (approved):
Good job! 🚀

PR #194, opened 2022-02-16
Hw solution

Sasha-Bachynskyi:


src/main/java/mate/jdbc/dao/ManufacturerDao.java
```diff
@@ -0,0 +1,19 @@
+package mate.jdbc.dao;
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Manufacturer;
+
+@Dao
```
LinkClink, line 8: `@Dao` antotation dont need here

src/main/java/mate/jdbc/util/ConnectionUtil.java
```diff
@@ -0,0 +1,28 @@
+package mate.jdbc.util;
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.util.Properties;
+
+public class ConnectionUtil {
+    static {
```
LinkClink, line 9: You can add constants to user, password, driver and connection link for more convenient editing of this data

LinkClink (changes requested):
look at this:

src/main/java/mate/jdbc/util/ConnectionUtil.java
LinkClink: only suggestion

LinkClink (approved):
Bomba

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,120 @@
...
+            getManufacturerStatement.setLong(INDEX_ID, id);
+            ResultSet resultSet = getManufacturerStatement.executeQuery();
+            if (!resultSet.next()) {
+                return Optional.empty();
+            }
+            manufacturer = getManufacturerFromResultSet(resultSet);
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get manufacturer from DB by id: " + id, e);
+        }
+        return Optional.ofNullable(manufacturer);
```
AlexRozv, line 62: if we change this "if" statement to the opposite (resultSet.next()) we can just return Optional of getManufacturerFromResultSet(resultSet)

and last return statement can just be Optional.empty()

this should simplify our code a little and make it more readable

AlexRozv (changes requested):

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
Sasha-Bachynskyi: agreed, it makes sense

AlexRozv (approved):
good job 👍

PR #201, opened 2022-03-06
Manufacturer dao was implemented

nastyaprokhorenko:


src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,118 @@
...
+                PreparedStatement getManufacturerStatement
+                        = connection.prepareStatement(getRequest)) {
+            getManufacturerStatement.setLong(1, id);
+            ResultSet resultSet = getManufacturerStatement.executeQuery();
+            if (resultSet.next()) {
+                String name = resultSet.getString("name");
+                String country = resultSet.getString("country");
+                manufacturer.setId(id);
+                manufacturer.setName(name);
+                manufacturer.setCountry(country);
```
MaliukDaria, line 54: checklist: "Try to avoid code duplication. Especially, when you are working with ResultSet. Move retrieving data from ResultSet into Entity object to a separate private method."

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,118 @@
...
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't insert manufacturer " + manufacturer
+                    + " to DB", e);
+        }
+        return manufacturer;
+    }
+
+    @Override
+    public Optional<Manufacturer> get(Long id) {
+        Manufacturer manufacturer = new Manufacturer();
```
MaliukDaria, line 42:
```suggestion
        Manufacturer manufacturer = null;
```

if there is no manufacturer with such id, we should return Optional.empty(), not new Manufacturer();

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,118 @@
...
+            getManufacturerStatement.setLong(1, id);
+            ResultSet resultSet = getManufacturerStatement.executeQuery();
+            if (resultSet.next()) {
+                String name = resultSet.getString("name");
+                String country = resultSet.getString("country");
+                manufacturer.setId(id);
+                manufacturer.setName(name);
+                manufacturer.setCountry(country);
+            }
+            return Optional.of(manufacturer);
```
MaliukDaria, line 56:
```suggestion
            return Optional.ofNullable(manufacturer);
```

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #205, opened 2022-03-10
First try

manglhorn:


src/main/java/dao/Manufacturer.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.Objects;
+
+public class Manufacturer {
+    private Long id;
+    private String name;
+    private String country;
+
+    public Manufacturer() {
+
+    }
```
MaliukDaria, line 12: why do we need this constructor?

src/main/java/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,118 @@
...
+            create.setString(1, manufacturer.getName());
+            create.setString(2, manufacturer.getCountry());
+            create.executeUpdate();
+            ResultSet generatorKeys = create.getGeneratedKeys();
+            if (generatorKeys.next()) {
+                Long id = generatorKeys.getObject(1, Long.class);
+                manufacturer.setId(id);
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't insert all manufacturers into DB", e);
```
MaliukDaria, line 34:
```suggestion
            throw new DataProcessingException("Can't insert manufacturer into DB " + manufacturer, e);

```

src/main/java/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,118 @@
...
+        return manufacturer;
+    }
+
+    @Override
+    public Optional<Manufacturer> get(Long id) {
+        String selectId = "SELECT * "
+                + "FROM manufacturers "
+                + "WHERE id = ? AND is_deleted = false;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement get = connection
```
MaliukDaria, line 45:
```suggestion
                PreparedStatement getManufacturer = connection

```

MaliukDaria (changes requested):

src/main/java/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,119 @@
...
+                    + manufacturer + " in DB ", e);
+        }
+    }
+
+    @Override
+    public boolean delete(Long id) {
+        String deleteRequest = "UPDATE manufacturers SET is_deleted = true WHERE id = ?;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement delete = connection
+                        .prepareStatement(deleteRequest, Statement.RETURN_GENERATED_KEYS)) {
```
KseniiaMakarova, line 101: why should we return generated keys here?

src/main/java/dao/Manufacturer.java
```diff
@@ -0,0 +1,65 @@
+package dao;
```
KseniiaMakarova, line 1: please do next HWs in the same package that is created (in this case, mate.jdbc)

KseniiaMakarova (approved):

KseniiaMakarova (commented):
also please be careful and don't do tasks in master branch, you should work in a separate branch

PR #235, opened 2022-05-05
Established connection to DB and implemented CRUD methods

Bukozdem:


src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,132 @@
...
+public class ManufacturerDaoImpl implements ManufacturerDao {
+    @Override
+    public List<Manufacturer> getAll() {
+        List<Manufacturer> allManufacturers = new ArrayList<>();
+        String getAllManufacturersRequest = "SELECT * FROM manufacturers WHERE is_deleted = false;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getAllManufacturersStatement =
+                        connection.prepareStatement(getAllManufacturersRequest)) {
+            ResultSet resultSet = getAllManufacturersStatement.executeQuery();
+            checkResultSet(resultSet);
```
MaliukDaria, line 26: I dont think we need this check

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,132 @@
...
+        String updateManufacturerRequest =
+                "UPDATE manufacturers SET name = ?, country = ? "
+                        + "WHERE is_deleted = false AND id = ?;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement updateManufacturerStatement =
+                        connection.prepareStatement(updateManufacturerRequest)) {
+            updateManufacturerStatement.setString(1, manufacturer.getName());
+            updateManufacturerStatement.setString(2, manufacturer.getCountry());
+            updateManufacturerStatement.setLong(3, manufacturer.getId());
+            Optional<Manufacturer> oldManufacturer = get(manufacturer.getId());
```
MaliukDaria, line 84: It's better not to open one connections in another, lets move this logic outside of try block

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,132 @@
...
+            getManufacturerStatement.setLong(1, id);
+            ResultSet resultSet = getManufacturerStatement.executeQuery();
+            checkResultSet(resultSet);
+            while (resultSet.next()) {
+                manufacturer = initializeManufacturer(resultSet);
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can`t get manufacturer from DB by id: " + id, e);
+        }
+        return Optional.of(manufacturer);
```
MaliukDaria, line 114:
```suggestion
        Manufacturer manufacturer = null;
        try (Connection connection = ConnectionUtil.getConnection();
                PreparedStatement getManufacturerStatement =
                        connection.prepareStatement(getManufacturerRequest)) {
            getManufacturerStatement.setLong(1, id);
            ResultSet resultSet = getManufacturerStatement.executeQuery();
            checkResultSet(resultSet);
            while (resultSet.next()) {
                manufacturer = initializeManufacturer(resultSet);
            }
        } catch (SQLException e) {
            throw new DataProcessingException("Can`t get manufacturer from DB by id: " + id, e);
        }
        return Optional.ofNullable(manufacturer);
```

as we should get empty Optional in case there is no manufacturer with such id

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,132 @@
...
+        }
+    }
+
+    private Manufacturer initializeManufacturer(ResultSet resultSet) {
+        try {
+            return new Manufacturer(resultSet.getObject("id", Long.class),
+                    resultSet.getString("name"),
+                    resultSet.getString("country"));
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can`t initialize manufacturer", e);
```
MaliukDaria, line 129:
```suggestion
            throw new DataProcessingException("Can`t get data from result set " + resultSet, e);
```

src/main/java/mate/jdbc/model/Manufacturer.java
```diff
@@ -0,0 +1,76 @@
...
+
+    public Manufacturer(String name, String country) {
+        this.name = name;
+        this.country = country;
+    }
+
+    public Manufacturer(Long id, String name, String country) {
+        this.id = id;
+        this.name = name;
+        this.country = country;
```
MaliukDaria, line 21:
```suggestion
        this(name,country);
```

It's better to use one constructor in another

MaliukDaria (changes requested):

PR #271, opened 2022-06-08
implemented ManufacturerDao

viktoriia-haraniuk:
- Established connection to Database
- Created init_db.sql file in src/main/resources folder
- Created Manufacturer model
- Created DAO layer for Manufacturer model with all required methods
- Called all CRUD methods in Main

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -1,7 +1,35 @@
 package mate.jdbc;
 
-public class Main {
-    public static void main(String[] args) {
+import mate.jdbc.dao.ManufacturerDao;
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Manufacturer;
 
+public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
+    public static void main(String[] args) {
+        ManufacturerDao manufacturerDao =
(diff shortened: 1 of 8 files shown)


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,35 @@
...
+
+        System.out.println(manufacturerDao.get(1L));
+        System.out.println(manufacturerDao.get(2L));
+
+        manufacturer.setId(1L);
+        manufacturer.setName("Opel");
+        manufacturer.setCountry("Germany");
+        manufacturerDao.update(manufacturer);
+
+        manufacturerDao.delete(3L);
```
boroda4436, line 31: Didn't you have an exception in this case? I see that you have saved only two instances to DB.

viktoriia-haraniuk: No, I didn't. I created one instace while I was implementing all dao's methods. So in total I had 3 record in db

viktoriia-haraniuk: image

viktoriia-haraniuk: NVM, my bad. I have just known about truncate 🤦‍♀️ 
I Improved my solution

boroda4436 (approved):

PR #279, opened 2022-06-16
implement task

gdpbalt:


src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,121 @@
...
+            statement.setString(2, manufacturer.getCountry());
+            statement.executeUpdate();
+
+            ResultSet resultSet = statement.getGeneratedKeys();
+            if (resultSet.next()) {
+                Long id = resultSet.getObject(1, Long.class);
+                manufacturer.setId(id);
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't create new record in DB. Query=" + query, e);
```
viktoriia-haraniuk, line 37: I think it is better to show information about manufacturer or any incoming data instead of Query message.
That would be more user friendly.
Like here 
image

viktoriia-haraniuk (changes requested):

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,122 @@
...
+            statement.setString(2, manufacturer.getCountry());
+            statement.executeUpdate();
+
+            ResultSet resultSet = statement.getGeneratedKeys();
+            if (resultSet.next()) {
+                Long id = resultSet.getObject(1, Long.class);
+                manufacturer.setId(id);
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Error create new record in DB", e);
```
KostyaNikiforov, line 37: Let's add info about manufacturer to the exception message

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,122 @@
...
+                Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement statement = connection.prepareStatement(query)
+        ) {
+            statement.setString(1, manufacturer.getName());
+            statement.setString(2, manufacturer.getCountry());
+            statement.setLong(3, manufacturer.getId());
+            statement.executeUpdate();
+        } catch (SQLException e) {
+            throw new DataProcessingException("Error update record in DB. ID="
+                    + manufacturer.getId(), e);
```
KostyaNikiforov, line 76: Any field can cause an error, so it is be cool if we see information about all of them in the message

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,122 @@
...
+        try (
+                Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement statement = connection.prepareStatement(query)
+        ) {
+            ResultSet resultSet = statement.executeQuery();
+            while (resultSet.next()) {
+                allManufacturers.add(createModelFromResult(resultSet));
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Error get records from DB", e);
```
KostyaNikiforov, line 109:
```suggestion
            throw new DataProcessingException("Error get all manufactories from DB", e);
```

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,41 @@
...
 
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.dao.ManufacturerDao;
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Manufacturer;
+
 public class Main {
+    private static final String PACKAGE_NAME = "mate.jdbc";
+    private static final Class<ManufacturerDao> CLASS_NAME = ManufacturerDao.class;
```
KostyaNikiforov, line 11: This constant seems unnecessary, cause there are can be a lot of injectorings and it would be strange to create s constant for each class.

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,41 @@
...
+
     public static void main(String[] args) {
+        Injector injector = Injector.getInstance(PACKAGE_NAME);
```
KostyaNikiforov, line 14: An opposite situation here, one injector object we can reuse a lot of times so it would be better to move it to the class level

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #305, opened 2022-07-04
-added my solution of JDBC

kqlynero:


nazarbaklanov (changes requested):
Хорошее решение, есть пару предложений.

nazarbaklanov (approved):
good solution

src/main/java/mate/jdbc/exceptions/DataProcessingException.java
```diff
@@ -0,0 +1,7 @@
+package mate.jdbc.exceptions;
```
dSkarzh, line 1:
```suggestion
package mate.jdbc.exception;
```
better use singular for package names

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,119 @@
...
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class ManufacturerDaoImpl implements ManufacturerDao {
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        String insertManufacturerRequest = "INSERT INTO manufacturers(name, country) values(?,?)";
+        try (Connection connection = ConnectionUtil.getConnection();
+                 PreparedStatement createManufacturerStatement = connection
```
dSkarzh, line 22:
```suggestion
        String query = "INSERT INTO manufacturers(name, country) values(?,?)";
        try (Connection connection = ConnectionUtil.getConnection();
                 PreparedStatement statement = connection
```
use simpler names for queries and statements everywhere, because it is obvious from the context what they are

dSkarzh (changes requested):

PR #321, opened 2022-08-11
created DB and implemented DAO logic

ssplot56:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,40 @@
...
+            = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
+        Manufacturer manufacturer1 = new Manufacturer();
+        manufacturer1.setName("Peugeot");
+        manufacturer1.setCountry("France");
+
+        Manufacturer manufacturer2 = new Manufacturer();
+        manufacturer2.setName("Volkswagen");
+        manufacturer2.setCountry("German");
```
MaliukDaria, line 18: It's bad practice to use numbers in variable names

MaliukDaria (approved):

PR #322, opened 2022-08-11
hw-jdbc

RomanPlk:


src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,113 @@
...
+            return Optional.ofNullable(manufacturer);
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get a manufacturer with id " + id, e);
+        }
+    }
+
+    @Override
+    public List<Manufacturer> getAll() {
+        List<Manufacturer> manufacturers = new ArrayList<>();
+        String getAllQuery = ("SELECT * FROM manufacturers WHERE is_deleted = false");
```
MaliukDaria, line 62:
```suggestion
        String getAllQuery = "SELECT * FROM manufacturers WHERE is_deleted = false";
```

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,113 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getManufacturersStatement = connection
+                        .prepareStatement(getAllQuery);) {
+            ResultSet resultSet = getManufacturersStatement
+                    .executeQuery(getAllQuery);
+            while (resultSet.next()) {
+                manufacturers.add(getManufacturer(resultSet));
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get all formats from DB", e);
```
MaliukDaria, line 72:
```suggestion
            throw new DataProcessingException("Can't get all manufacturers from DB", e);
```

src/main/java/mate/jdbc/model/Manufacturer.java
```diff
@@ -0,0 +1,51 @@
...
+
+    public Manufacturer(String name, String country) {
+        this.name = name;
+        this.country = country;
+    }
+
+    public Manufacturer(Long id, String name, String country) {
+        this.id = id;
+        this.name = name;
+        this.country = country;
```
MaliukDaria, line 16:
```suggestion
        this(name, country);
```

It is better to use one constructor into another

MaliukDaria (changes requested):

aaananas (approved):

PR #337, opened 2022-09-06
connected to database, created model Manufacturer and dao Manufacture…

semwithuncle:
…rDao

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -1,7 +1,43 @@
 package mate.jdbc;
 
-public class Main {
-    public static void main(String[] args) {
+import java.util.Arrays;
+import java.util.Optional;
+import mate.jdbc.dao.ManufacturerDao;
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Manufacturer;
 
+public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
+    public static void main(String[] args) {
+        ManufacturerDao manufacturerDao =
+                (ManufacturerDao) injector.getInstance(ManufacturerDao.class);
+        Manufacturer peugeot = new Manufacturer(null, "Peugeot", "France");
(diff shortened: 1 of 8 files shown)


src/main/java/mate/jdbc/util/ConnectionUtil.java
```diff
@@ -0,0 +1,33 @@
+package mate.jdbc.util;
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.util.Properties;
+import mate.jdbc.exception.DataProcessingException;
+
+public class ConnectionUtil {
+    public static final String USER = "user";
```
GolDenik, line 10: this is redundant
```suggestion
```

src/main/java/mate/jdbc/util/ConnectionUtil.java
```diff
@@ -0,0 +1,33 @@
...
+    public static final String USER_NAME = "root";
+    public static final String PASSWORD = "password";
```
GolDenik, line 12: I think this is redundant
```suggestion
```

GolDenik (changes requested):

GolDenik (approved):

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,43 @@
...
+        System.out.println(manufacturerGetRenault);
+
+        boolean delete = manufacturerDao.delete(1L);
+        boolean deleteRenault = manufacturerDao.delete(2L);
+        System.out.println(delete);
+        System.out.println(deleteRenault);
+        System.out.println(manufacturerDao.getAll());
+
+        manufacturerDao.update(new Manufacturer(3L, "Zaz", "Ukraine"));
+        System.out.println(manufacturerDao.get(3L));
```
Noct2000, line 39: don't use indexes especially for your DB, use indexes that definitely will be in DB after running your code

Noct2000 (changes requested):

Noct2000 (approved):

PR #381, opened 2022-09-09
JDBS Intro manufacturers DAO impl v1

EvgenyTyshchenko:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,30 @@
...
+        System.out.println("Created manufacturer: " + manufacturer);
+        manufacturer.setName("New name");
+        manufacturer.setCountry("New country");
+        manufacturerDao.update(manufacturer);
+        System.out.println("Updated manufacturer: " + manufacturer);
+        Optional<Manufacturer> manufacturer1 = manufacturerDao.get(manufacturer.getId());
+        System.out.println("Manufacturer by code 1: " + manufacturer1);
+        manufacturerDao.delete(manufacturer.getId());
+        manufacturerDao.getAll().forEach(System.out::println);
+        Optional<Manufacturer> manufacturer2 = manufacturerDao.get(12345L);
```
liliia-ponomarenko, line 27: It’s  bad practice to name variable with counter -> ```manufacturer1, manufacturer2```. Also, it is better to leave only really important tent comments (now it’s not critical but use this advice when you will write your project).

liliia-ponomarenko (changes requested):
Good job 🚀  Let’s improve your solution😉

liliia-ponomarenko (approved):
good job! 👍

PR #402, opened 2022-10-24
solution was added

Kostiantyn-Sacheiev:


src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,149 @@
...
+            "INSERT INTO manufacturers (name, country) VALUES (?, ?);";
+    private static final String GET_MANUFACTURER_REQUEST =
+            "SELECT * FROM manufacturers WHERE id = ?;";
+    private static final String GET_ALL_MANUFACTURERS_REQUEST =
+            "SELECT * FROM manufacturers WHERE is_deleted = FALSE";
+    private static final String UPDATE_REQUEST =
+            "UPDATE manufacturers SET name = ?, country = ? WHERE id = ?;";
+    private static final String DELETE_REQUEST =
+            "UPDATE manufacturers SET is_deleted = true where id = ?;";
+    private static final Connection CONNECTION = ConnectionUtil.getConnection();
```
aaananas, line 28: its is a bad practice to use constants for SQL requests - your code is hard to read
also there is not need to constant for connection either

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,149 @@
...
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        try (PreparedStatement createManufacturerStatement =
+                CONNECTION.prepareStatement(INSERT_MANUFACTURER_REQUEST,
+                        Statement.RETURN_GENERATED_KEYS)) {
+            createNewManufacturer(manufacturer, createManufacturerStatement);
+            createKey(manufacturer, createManufacturerStatement);
```
aaananas, line 36: all those methods make your code hard to read
`extractManufacturerFromResultSet` will be enough

aaananas (changes requested):

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,148 @@
...
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        String insertManufacturerRequest =
+                "INSERT INTO manufacturers (name, country) VALUES (?, ?);";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement createManufacturerStatement =
+                        connection.prepareStatement(insertManufacturerRequest,
+                        Statement.RETURN_GENERATED_KEYS)) {
+            extractManufacturerFromResultSet(manufacturer, createManufacturerStatement);
```
zachar-chorny, line 27: Don't put this logic in a separate method, it will compromise the readability of the method. It also does not make sense because this code is not repeated.

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,148 @@
...
+                "SELECT * FROM manufacturers WHERE id = ?;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getManufacturerStatement =
+                        connection.prepareStatement(getManufacturerRequest,
+                        Statement.NO_GENERATED_KEYS)) {
+            getManufacturerStatement.setLong(1, id);
+            ResultSet manufacturerResultSet = getResultSet(getManufacturerStatement);
+            List<Manufacturer> manufacturersList =
+                    convertResultSetToManufacturersList(manufacturerResultSet);
+            return findManufacturerOnId(id, manufacturersList);
```
zachar-chorny, line 46: Why make things so complicated? Do it all at once here, you don't need so many private methods

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,148 @@
...
+    @Override
+    public List<Manufacturer> getAll() {
+        String getAllManufacturersRequest =
+                "SELECT * FROM manufacturers WHERE is_deleted = FALSE";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getAllManufacturersStatement =
+                        connection.prepareStatement(getAllManufacturersRequest,
+                             Statement.NO_GENERATED_KEYS)) {
+            ResultSet manufacturerResultSet = getResultSet(getAllManufacturersStatement);
+            return convertResultSetToManufacturersList(manufacturerResultSet);
```
zachar-chorny, line 61: Same thing, it makes no sense to use the convertResultSetToManufacturersList method in the getAll, so you don't need it at all, write this code right here in the getAll

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,148 @@
...
+        createManufacturerStatement.setString(2, manufacturer.getCountry());
+        createManufacturerStatement.executeUpdate();
+        ResultSet generatedKeys = createManufacturerStatement.getGeneratedKeys();
+        if (generatedKeys.next()) {
+            Long id = generatedKeys.getObject(1, Long.class);
+            manufacturer.setId(id);
+        }
+    }
+
+    private ResultSet getResultSet(PreparedStatement preparedStatement) throws SQLException {
```
zachar-chorny, line 110: Please remove all private methods except convertResultSetToManufacturersList only change its return type to Manufacturer instead of list. Then you can call it in getAll() just in loop. And delete other private methods, and write their code right where you call them.  Remember, only repeatable code should be included in private methods, or if the amount of code makes the method unreadable then it makes sense to put it all in private methods, but this is not your case

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    @Override
+    public Manufacturer update(Manufacturer manufacturer) {
+        String updateRequest =
+                "UPDATE manufacturers SET name = ?, country = ? WHERE id = ?;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement updateManufacturerStatement =
+                        connection.prepareStatement(updateRequest, Statement.NO_GENERATED_KEYS)) {
+            get(manufacturer.getId());
```
aaananas, line 87: why do you get a manufacturer here?

aaananas (approved):

PR #403, opened 2022-10-26
taxi service db is connected

mrmax24:


kateryna-deshchinska (changes requested):
You also need to create init_db.sql, read your task description again

Rommelua (approved):

PR #404, opened 2022-10-27
First solution

Mariana-VV:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,36 @@
...
+        //      System.out.println(manufacturerDao.delete(1l));
+
+        //      Manufacturer manufacturer1 = new Manufacturer(5,"maibach","DUBAI");
+        //      System.out.println(manufacturerDao.update(manufacturer1));
 
+        //      System.out.println(manufacturerDao.get(5l));
+        //
+        //      List<Manufacturer> manufacturers = manufacturerDao.getAll();
+        //      for (Manufacturer manufacturer : manufacturers) {
+        //      System.out.println(manufacturer);     }
```
zachar-chorny, line 34: First of all, why do you need an object of the ConnectionUtil class? And secondly, why have you commented out everything?

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,110 @@
...
+import mate.jdbc.model.DataProcessingException;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class ManufacturerDaoImpl extends ConnectionUtil implements ManufacturerDao {
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        String sql = "insert into manufacturers(name,country) values("
```
zachar-chorny, line 20:
```suggestion
        String query = "insert into manufacturers(name,country) values("
```

zachar-chorny (changes requested):

PR #410, opened 2022-11-02
implemented Manufacturer and ManufacturerDao classes

nazariipertsak:


src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,118 @@
...
+                         connection.prepareStatement(selectAllRequest);) {
+            ResultSet resultSet = getAllManufacturers.executeQuery();
+            while (resultSet.next()) {
+                String name = resultSet.getString("name");
+                String country = resultSet.getString("country");
+                Long id = resultSet.getObject("id", Long.class);
+                Manufacturer manufacturer = new Manufacturer();
+                manufacturer.setName(name);
+                manufacturer.setCountry(country);
+                manufacturer.setId(id);
```
Pavlo-Gook, line 77: Perhaps this part could be moved out of get() and getAll() into a separate method. My recommendation

src/main/resources/init_db.sql
```diff
@@ -0,0 +1,9 @@
+CREATE DATABASE `taxi_service_db` /*!40100 DEFAULT CHARACTER SET utf8mb3 */ /*!80016 DEFAULT ENCRYPTION='N' */;
```
ChumakAn, line 1:
```suggestion
CREATE DATABASE IF NOT EXISTS `taxi_service_db`;
```

src/main/resources/init_db.sql
```diff
@@ -0,0 +1,9 @@
...
+
+CREATE TABLE `manufacturers` (
+  `id` bigint NOT NULL AUTO_INCREMENT,
+  `name` varchar(255) NOT NULL,
+  `country` varchar(255) NOT NULL,
+  `is_deleted` tinyint NOT NULL DEFAULT '0',
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB AUTO_INCREMENT=3 DEFAULT CHARSET=utf8mb3;
```
ChumakAn, line 9: why AUTO_INCREMENT=3?

ChumakAn (changes requested):

src/main/resources/init_db.sql
nazariipertsak: Because at moment when i copied create statement alredy was 3 elements in database. I will change that

ChumakAn (approved):

PR #411, opened 2022-11-02
ManufacturerDao implemented

hpavlo:


nazariipertsak (approved):
LGTM!

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,129 @@
...
+    private Optional<Manufacturer> getEntity(ResultSet resultSet) throws SQLException {
+        Long id = resultSet.getObject(ID_COLUMN, Long.class);
+        String name = resultSet.getString(NAME_COLUMN);
+        String country = resultSet.getString(COUNTRY_COLUMN);
+        boolean isDeleted = resultSet.getBoolean(IS_DELETED_COLUMN);
+        if (!isDeleted) {
+            return Optional.of(new Manufacturer(id, name, country));
+        }
+        return Optional.empty();
+    }
```
KostyaNikiforov, line 128: Interesting solution, but there is no need to fetch data from db if you are not going to show it to the user. Let's add a filtration to the query

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,129 @@
...
+    private static final String CREATE_MANUFACTURER_REQUEST =
+            "INSERT INTO manufacturers(name, country) VALUES(?, ?);";
+    private static final String GET_MANUFACTURER_REQUEST =
+            "SELECT * FROM manufacturers WHERE id = ?;";
+    private static final String GET_ALL_MANUFACTURERS_REQUEST =
+            "SELECT * FROM manufacturers;";
+    private static final String UPDATE_MANUFACTURER_REQUEST =
+            "UPDATE manufacturers SET name = ?, country = ? WHERE id = ?;";
+    private static final String DELETE_MANUFACTURER_REQUEST =
+            "UPDATE manufacturers SET is_deleted = true WHERE id = ?;";
```
KostyaNikiforov, line 32: Constant queries are redundant. You can think about it as about a code, and you can write them inside of the method

KostyaNikiforov (changes requested):

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,124 @@
...
+
+@Dao
+public class ManufacturerDaoImpl implements ManufacturerDao {
+    private static final int ID_COLUMN = 1;
+    private static final int NAME_COLUMN = 2;
+    private static final int COUNTRY_COLUMN = 3;
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        String createManufacturerRequest =
```
KostyaNikiforov, line 25: It would be enough to just name it `query`, because we already see the method signature

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,124 @@
...
+            throw new DataProcessingException("Can't insert manufacturer='"
+                    + manufacturer + "' to DB", e);
+        }
+        return manufacturer;
+    }
+
+    @Override
+    public Optional<Manufacturer> get(Long id) {
+        String getManufacturerRequest =
+                "SELECT * FROM manufacturers WHERE id = ? AND is_deleted = 0;";
```
KostyaNikiforov, line 49:
```suggestion
                "SELECT * FROM manufacturers WHERE id = ? AND is_deleted = FALSE;";
```

KostyaNikiforov (changes requested):

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
KostyaNikiforov: Screenshot 2022-11-03 at 11 13 39

KostyaNikiforov (approved):

liliia-ponomarenko (approved):

PR #417, opened 2022-11-05
implemented task

vov4ik89:


src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,113 @@
...
+            throw new DataProcessingException("Can't update manufacturer " + manufacturer, e);
+        }
+    }
+
+    @Override
+    public boolean delete(Long id) {
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement deleteStatement = connection.prepareStatement(DELETE_REQUEST)) {
+            deleteStatement.setLong(1, id);
+            return deleteStatement.executeUpdate() >= 1;
```
Andew-Miroshnikov, line 100: return deleteStatement.executeUpdate() > 0;

Andew-Miroshnikov (changes requested):

Andew-Miroshnikov (approved):

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,113 @@
...
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get manufacturer from DB by id " + id, e);
+        }
+        return Optional.empty();
+    }
+
+    @Override
+    public List<Manufacturer> getAll() {
+        List<Manufacturer> allManufacturers = new ArrayList<>();
```
hronograf, line 67: it would be better to move declaration of `allManufacturers` variable as close to its usage as possible
so it should be near `if` at 70th line
check this in other places

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,113 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                Statement getAllStatement = connection.createStatement()) {
```
hronograf, line 69: image

hronograf (changes requested):
Good!
also, it would be better to have separate package for exceptions

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -41,47 +32,51 @@ public Manufacturer create(Manufacturer manufacturer) {
...
         }
-        return manufacturer;
     }
 
     @Override
     public Optional<Manufacturer> get(Long id) {
+        String query = "SELECT * FROM manufacturers WHERE id = ? AND is_deleted = false";
         try (Connection connection = ConnectionUtil.getConnection();
-                PreparedStatement getStatement = connection.prepareStatement(GET_BY_ID_REQUEST)) {
+                PreparedStatement getStatement = connection.prepareStatement(query)) {
```
hronograf, line 45: we can also write just `statement` instead of `getStatement`
like we wrote `query` and `connection`

hronograf (approved):
Great! 👍

PR #418, opened 2022-11-06
Create connection to DB and implement methods in DAO

Andew-Miroshnikov:


src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,122 @@
...
+
+    private Manufacturer parse(ResultSet resultSet) {
+        Manufacturer manufacturer = new Manufacturer();
+        try {
+            manufacturer.setId(resultSet.getObject("id", Long.class));
+            manufacturer.setName(resultSet.getString("name"));
+            manufacturer.setCountry(resultSet.getString("country"));
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't parse data from ResultSet: "
+                    + resultSet, e);
```
vov4ik89, line 118: I think better rename method to getManufacturer and throw exeption in method - private Manufacturer getManufacturer(ResultSet resultSet) throws SQLException {

src/main/java/mate/jdbc/exception/DataProcessingException.java
```diff
@@ -0,0 +1,8 @@
+package mate.jdbc.exception;
+
+public class DataProcessingException extends RuntimeException {
+
```
vov4ik89, line 4: empty line

vov4ik89 (changes requested):
Good, only small changes

vov4ik89 (approved):

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,117 @@
...
+import mate.jdbc.exception.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.models.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class ManufacturerDaoImpl implements ManufacturerDao {
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        String insertManufacturerRequest = "INSERT INTO manufacturers(name, country) values(?, ?);";
```
hronograf, line 20: we can also name this variable just `query` instead of `insertManufacturerRequest`
because there is only one query and we see that this query inside `create` method

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,117 @@
...
+                    + manufacturer + " to DB", e);
+        }
+        return manufacturer;
+    }
+
+    @Override
+    public Optional<Manufacturer> get(Long id) {
+        String getByIdRequest = "SELECT * FROM manufacturers "
+                + "where is_deleted = false AND id = ?;";
+        Manufacturer manufacturer = null;
```
hronograf, line 44: it would be better to move declaration of `manufacturer` variable as close to its usage as possible
so it should be near `if` at 50th line

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,117 @@
...
+            getByIdManufacturerStatement.setLong(1, id);
+            ResultSet resultSet = getByIdManufacturerStatement.executeQuery();
+            if (resultSet.next()) {
+                manufacturer = getManufacturer(resultSet);
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get manufacturer from DB by id "
+                    + id, e);
+        }
+        return Optional.ofNullable(manufacturer);
```
hronograf, line 57: similar about `return`
we should return something just when it is possible
so `return` should be after `if` at 52nd line

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    @Override
+    public List<Manufacturer> getAll() {
+        List<Manufacturer> allManufacturers = new ArrayList<>();
```
hronograf, line 62: same

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,117 @@
...
+                        connection.prepareStatement(readRequest)) {
+            ResultSet resultSet = getAllManufacturersStatement
+                    .executeQuery();
+            while (resultSet.next()) {
+                allManufacturers.add(getManufacturer(resultSet));
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get all manufacturers from DB", e);
+        }
+        return allManufacturers;
```
hronograf, line 75: same
check this everywhere

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,117 @@
...
+            throw new DataProcessingException("Can't get manufacturer from DB by id "
+                    + id, e);
+        }
+        return Optional.ofNullable(manufacturer);
+    }
+
+    @Override
+    public List<Manufacturer> getAll() {
+        List<Manufacturer> allManufacturers = new ArrayList<>();
+        String readRequest = "SELECT * FROM manufacturers where is_deleted = false;";
```
hronograf, line 63: another possible variant of query (optional)
```suggestion
        String readRequest = "SELECT * FROM manufacturers where !is_deleted;";

```

hronograf (changes requested):
Good!

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -30,75 +30,75 @@ public Manufacturer create(Manufacturer manufacturer) {
...
-        return Optional.ofNullable(manufacturer);
     }
 
     @Override
     public List<Manufacturer> getAll() {
-        List<Manufacturer> allManufacturers = new ArrayList<>();
-        String readRequest = "SELECT * FROM manufacturers where is_deleted = false;";
+        String query = "SELECT * FROM manufacturers where !is_deleted;";
         try (Connection connection = ConnectionUtil.getConnection();
                 PreparedStatement getAllManufacturersStatement =
```
hronograf, line 64: we could name `getAllManufacturersStatement` just `statement` or `preparedStatement`

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -30,75 +30,75 @@ public Manufacturer create(Manufacturer manufacturer) {
...
-                        connection.prepareStatement(readRequest)) {
+                        connection.prepareStatement(query)) {
+            List<Manufacturer> allManufacturers = new ArrayList<>();
```
hronograf, line 66: `allManufacturers` could be even closer to the while loop

hronograf (approved):
Nice!

PR #424, opened 2022-11-09
Created model Manufacturer, ManufacturerDao and its implementation

PavloPolovyi:


alex-anofriev (approved):

src/main/java/mate/jdbc/dao/implementation/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,117 @@
...
+                PreparedStatement deleteStatement = connection.prepareStatement(sqlQuery)) {
+            deleteStatement.setLong(1, id);
+            int updatedRows = deleteStatement.executeUpdate();
+            return updatedRows > 0;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't delete manufacturer with id: " + id, e);
+        }
+    }
+
+    private Manufacturer processRow(ResultSet resultSet) throws SQLException {
```
liliia-ponomarenko, line 110: If you use try-catch you can show that an exception occurred when receiving data from ResultSet.

liliia-ponomarenko (approved):
great! 👍

PR #426, opened 2022-11-09
solution

vyshnevsky:


src/main/java/mate/jdbc/model/Manufacturer.java
```diff
@@ -0,0 +1,71 @@
...
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        Manufacturer that = (Manufacturer) o;
+        return Objects.equals(id, that.id) && Objects.equals(name, that.name)
+                && Objects.equals(country, that.country);
```
alex-anofriev, line 55: I am not sure if it is right to add here "id", because this way every object will  return false cause every id unique. Same with hashcode.

alex-anofriev (changes requested):

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,102 @@
...
+                     ConnectionUtil.getConnection().prepareStatement(query,
+                             Statement.RETURN_GENERATED_KEYS)) {
+            statement.setLong(1, id);
+            return statement.executeUpdate() >= 1;
+        } catch (SQLException ex) {
+            throw new DataProcessingException("Can't delete manufacturer in DB with id: " + id, ex);
+        }
+    }
+
+    private Manufacturer parseManufacturer(ResultSet resultSet) throws SQLException {
```
liliia-ponomarenko, line 96: If you use try-catch you can show that an exception occurred when receiving data from ResultSet.

src/main/resources/init_db.sql
```diff
@@ -0,0 +1,10 @@
+CREATE DATABASE `taxi_service_db` /*!40100 DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_ai_ci */ /*!80016 DEFAULT ENCRYPTION='N' */;
```
liliia-ponomarenko, line 1:
```suggestion
CREATE DATABASE `taxi_service_db`;
```

src/main/resources/init_db.sql
```diff
@@ -0,0 +1,10 @@
...
+
+CREATE TABLE `manufacturers` (
+  `id` int NOT NULL AUTO_INCREMENT,
+  `name` varchar(255) NOT NULL,
+  `country` varchar(255) NOT NULL,
+  `is_deleted` tinyint NOT NULL DEFAULT '0',
+  PRIMARY KEY (`id`),
+  UNIQUE KEY `id_UNIQUE` (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;
```
liliia-ponomarenko, line 10:
```suggestion
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
```
It’s  unnecessary  because  ```id``` AUTO_INCREMENT

liliia-ponomarenko (changes requested):
great! Let’s fix a few mistakes ;)

liliia-ponomarenko (approved):
LGTM 🔥

PR #438, opened 2022-11-13
Realized simple CRUD operations.

OleksandrKorniychuk:


src/main/java/mate/jdbc/model/Manufacturer.java
```diff
@@ -0,0 +1,76 @@
+package mate.jdbc.model;
+
+import java.util.Objects;
+
+public class Manufacturer {
+    private Long id;
+    private String name;
+    private String country;
+
+    public Manufacturer() {
```
romangolovatyi, line 10: I think you don't need clearly announce an empty constructor if you don't have another constructors.

romangolovatyi (approved):
Great!!

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,125 @@
...
+            throw new DataProcessingException("Can't insert manufacturer to DB", e);
+        }
+        return manufacturer;
+    }
+
+    @Override
+    public Optional<Manufacturer> get(Long id) {
+        String query = "SELECT * "
+                + "FROM manufacturers "
+                + "WHERE id = ? AND is_deleted = false;";
```
ChumakAn, line 48: why do you need to concat strings here?

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,125 @@
...
+            throw new DataProcessingException("Can't get manufacturer from DB by id = " + id, e);
+        }
+        return optionalManufacturer;
+    }
+
+    @Override
+    public List<Manufacturer> getAll() {
+        String query = "SELECT * "
+                + "FROM manufacturers "
+                + "WHERE is_deleted = false;";
```
ChumakAn, line 68: same for all, it's redundant

src/main/java/mate/jdbc/model/Manufacturer.java
```diff
@@ -0,0 +1,76 @@
...
+
```
ChumakAn, line 11: empty line

ChumakAn (changes requested):
Good! Let's make it better

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
OleksandrKorniychuk: I am used to using SQL queries from new line...

ChumakAn (approved):

PR #444, opened 2022-11-14
Hw jv jdbc solution

romangolovatyi:


OleksandrKorniychuk (approved):
Looks good! 🚀
But i think better move this: 
src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
to `impl` folder:
src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,117 @@
...
+import mate.jdbc.exceptions.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.models.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class ManufacturerDaoImpl implements ManufacturerDao {
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        String insertManufacturerRequest = "INSERT INTO manufacturers(name, country) values(?, ?);";
```
ChumakAn, line 21:
```suggestion
        String query = "INSERT INTO manufacturers(name, country) values(?, ?);";
```

ChumakAn: `query` is enough

ChumakAn: same for all

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,117 @@
...
+    public Manufacturer update(Manufacturer manufacturer) {
+        String updateManufacturerRequest = "UPDATE manufacturers SET name = ?, country = ? "
+                + "WHERE is_deleted = FALSE AND id = ?";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement preparedStatement = connection
+                        .prepareStatement(updateManufacturerRequest)) {
+            preparedStatement.setString(1, manufacturer.getName());
+            preparedStatement.setString(2, manufacturer.getCountry());
+            preparedStatement.setObject(3, manufacturer.getId());
+            preparedStatement.executeUpdate();
```
ChumakAn, line 87: let's check if the manufacturer with such id exists before executing update

src/main/java/mate/jdbc/util/ConnectionUtil.java
```diff
@@ -0,0 +1,28 @@
...
+            throw new RuntimeException("Can't load JDBC driver for MySQL", e);
+        }
+    }
+
+    public static Connection getConnection() {
+        try {
+            Properties dbProperties = new Properties();
+            dbProperties.put("user", "root");
+            dbProperties.put("password", "12345678");
+            return DriverManager.getConnection("jdbc:mysql://localhost:3306/taxi_service",
```
ChumakAn, line 22: it's better to make the credentials and driver name a constants

ChumakAn (changes requested):
Good, let's make it better

src/main/java/mate/jdbc/util/ConnectionUtil.java
romangolovatyi: Thank you for your remarks

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
ChumakAn: just an idea, not mandatory

ChumakAn (approved):

PR #461, opened 2022-12-26
Added ManufacturerDao interface with some methods and added implement…

AntonKalenskiy:
…ation ManufacturerDaoImpl. Created own exception, added model Manufacturer and added

 init_db.sql file.

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -1,7 +1,18 @@
 package mate.jdbc;
 
-public class Main {
-    public static void main(String[] args) {
+import mate.jdbc.dao.ManufacturerDao;
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Manufacturer;
 
+public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
+    public static void main(String[] args) {
+        ManufacturerDao manufacturerDao =
+                (ManufacturerDao) injector.getInstance(ManufacturerDao.class);
+        Manufacturer manufacturer = new Manufacturer();
+        manufacturer.setName("Enrike");
+        manufacturer.setCountry("Brazil");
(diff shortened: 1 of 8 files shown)


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,18 @@
...
-    public static void main(String[] args) {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
 
+    public static void main(String[] args) {
+        ManufacturerDao manufacturerDao =
+                (ManufacturerDao) injector.getInstance(ManufacturerDao.class);
+        Manufacturer manufacturer = new Manufacturer();
+        manufacturer.setName("Enrike");
+        manufacturer.setCountry("Brazil");
+        manufacturerDao.create(manufacturer);
```
aleksandr-bevza, line 16: You should check all manufacturerDao methods here, not only "create"

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,122 @@
...
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class ManufacturerDaoImpl implements ManufacturerDao {
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        String insertManufacturerRequest = "INSERT INTO manufacturers(name, country) VALUES(?, ?);";
```
aleksandr-bevza, line 21: For SQL it better to use word "query" instead of "request".
Fix it in other places.
```suggestion
        String insertManufacturerQuery = "INSERT INTO manufacturers(name, country) VALUES(?, ?);";
```

aleksandr-bevza (changes requested):

nacenik (approved):
Good job!

PR #487, opened 2023-01-21
JDBC - 1st commit

OleksandrKostiukevych:


MaxMaksym (changes requested):

src/main/java/mate/jdbc/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,114 @@
...
+                        connection.prepareStatement(query)) {
+            deleteManufacturerStatement.setLong(1, id);
+            return deleteManufacturerStatement.executeUpdate() > 0;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't delete manufacturer "
+                    + "from database by id:" + id + ".", e);
+        }
+    }
+
+    private Optional<Manufacturer> getManufacturer(ResultSet resultSet) throws SQLException {
```
MaxMaksym, line 108:
```suggestion
    private Manufacturer getManufacturer(ResultSet resultSet) throws SQLException {
```
just return manufacturer, you  don't need to use Optional here

src/main/java/mate/jdbc/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,114 @@
...
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get all manufacturers from database.", e);
+        }
+        return manufacturers;
+    }
+
+    @Override
+    public Manufacturer update(Manufacturer manufacturer) {
+        String query = "UPDATE manufacturers SET name = ?, country = ? WHERE id = ?;";
```
MaxMaksym, line 80:
```suggestion
        String query = "UPDATE manufacturers SET name = ?, country = ? WHERE id = ? AND is_deleted = FALSE";
```

MaxMaksym (approved):

src/main/java/mate/jdbc/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,114 @@
...
+import mate.jdbc.lib.Dao;
+import mate.jdbc.lib.exception.DataProcessingException;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class ManufacturerDaoImpl implements ManufacturerDao {
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        String query = "INSERT INTO manufacturers(name, country) VALUES(?, ?);";
```
KostyaNikiforov, line 21:
```suggestion
        String query = "INSERT INTO manufacturers(name, country) VALUES(?, ?)";
```

src/main/java/mate/jdbc/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,114 @@
...
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't insert manufacturer: "
+                    + manufacturer + " to database.", e);
+        }
+        return manufacturer;
+    }
+
+    @Override
+    public Optional<Manufacturer> get(Long id) {
+        String query = "SELECT * FROM manufacturers WHERE id = ? AND is_deleted = false;";
```
KostyaNikiforov, line 43: Fix it everywhere
```suggestion
        String query = "SELECT * FROM manufacturers WHERE id = ? AND is_deleted = false";
```

src/main/java/mate/jdbc/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,115 @@
...
+        }
+        return manufacturer;
+    }
+
+    @Override
+    public Optional<Manufacturer> get(Long id) {
+        String query = "SELECT * FROM manufacturers WHERE id = ? AND is_deleted = false";
+        Optional<Manufacturer> optionalManufacturer = Optional.empty();
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getManufacturerStatement =
```
KostyaNikiforov, line 46:
```suggestion
                PreparedStatement statement =
```

src/main/java/mate/jdbc/impl/ManufacturerDaoImpl.java
KostyaNikiforov: Agree, this method never return null

KostyaNikiforov (changes requested):

src/main/java/mate/jdbc/impl/ManufacturerDaoImpl.java
```diff
@@ -96,20 +95,20 @@ public Manufacturer update(Manufacturer manufacturer) {
...
+            return statement.executeUpdate() > 0;
         } catch (SQLException e) {
             throw new DataProcessingException("Can't delete manufacturer "
                     + "from database by id:" + id + ".", e);
         }
     }
 
-    private Optional<Manufacturer> getManufacturer(ResultSet resultSet) throws SQLException {
-        return Optional.of(new Manufacturer(
+    private Manufacturer getManufacturer(ResultSet resultSet) throws SQLException {
```
KostyaNikiforov, line 108: No. This method will never return null, so you don't need to return an Optional here

src/main/java/mate/jdbc/impl/ManufacturerDaoImpl.java
```diff
@@ -39,35 +38,35 @@ public Manufacturer create(Manufacturer manufacturer) {
     }
 
     @Override
-    public Optional<Manufacturer> get(Long id) {
```
KostyaNikiforov: But this method can return a null, if there are no any manufacturer with such id. So here you have to return an Optional

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

src/main/java/mate/jdbc/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,115 @@
...
+            throw new DataProcessingException("Can't insert manufacturer: "
+                    + manufacturer + " to database.", e);
+        }
+        return manufacturer;
+    }
+
+    @Override
+    public Optional<Manufacturer> get(Long id) {
+        String query = "SELECT * FROM manufacturers WHERE id = ? AND is_deleted = false";
+        Manufacturer manufacturer = new Manufacturer();
```
KostyaNikiforov, line 44:
```suggestion
        Manufacturer manufacturer = null;
```

src/main/java/mate/jdbc/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,115 @@
...
+            statement.setLong(1, id);
+            ResultSet resultSet = statement.executeQuery();
+            if (resultSet.next()) {
+                manufacturer = getManufacturer(resultSet);
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get manufacturer from "
+                    + "database by id: " + id + ".", e);
+        }
+        return Optional.of(manufacturer);
```
KostyaNikiforov, line 57:
```suggestion
        return Optional.ofNullable(manufacturer);
```

KostyaNikiforov (approved):

PR #488, opened 2023-01-21
jdbc created

solomiia-voloshyn:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,36 @@
...
+
 public class Main {
-    public static void main(String[] args) {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
 
+    public static void main(String[] args) {
+        ManufacturerDao manufacturerDao =
+                (ManufacturerDao) injector.getInstance(ManufacturerDao.class);
+        Manufacturer manufacturerKia =
+                new Manufacturer("KIA", "Japan");
```
Avassor3000, line 15: i guess kia is from korea)

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,111 @@
...
+                PreparedStatement statement = connection.prepareStatement(query)) {
+            return getManufacturersFromResultSet(statement);
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get all manufacturers from DB ", e);
+        }
+    }
+
+    @Override
+    public Manufacturer update(Manufacturer manufacturer) {
+        String query = "UPDATE manufacturers SET country = ? WHERE name = ?;";
```
Avassor3000, line 69: i guess we have to update not only country but name too

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,111 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement statement = connection.prepareStatement(query)) {
+            statement.setString(FIRST_PARAMETER, manufacturer.getCountry());
+            statement.setString(SECOND_PARAMETER, manufacturer.getName());
+            statement.executeUpdate();
+            return manufacturer;
```
Avassor3000, line 75: we have to return manufacturer if invoking of update would be grater than 0

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
Avassor3000, line 69: i guess we have to update not only country but name too

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
Avassor3000, line 75: we have to return manufacturer if invoking of update would be grater than 0

Avassor3000 (changes requested):

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
solomiia-voloshyn: done

Avassor3000 (approved):

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.NoSuchElementException;
+import java.util.Optional;
+import mate.jdbc.exception.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class ManufacturerDaoImpl implements ManufacturerDao {
+    private static final int FIRST_ELEMENT_IN_LIST = 0;
```
KostyaNikiforov, line 19: int is not an element, int can be an index) And i find this constant redundant

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,115 @@
...
+            return manufacturer;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't insert manufacturer to DB. Manufacturer name: "
+                    + manufacturer.getName(), e);
+        }
+    }
+
+    @Override
+    public Optional<Manufacturer> get(Long id) {
+        String query = "SELECT * FROM manufacturers WHERE id = ? AND is_deleted = false;";
```
KostyaNikiforov, line 44:
```suggestion
        String query = "SELECT * FROM manufacturers WHERE id = ? AND is_deleted = false";

```

src/main/java/mate/jdbc/util/ConnectionUtil.java
```diff
@@ -0,0 +1,27 @@
...
+import java.sql.SQLException;
+
+public class ConnectionUtil {
+    private static final String DB_PASSWORD = "1234";
+    private static final String DB_URL = "jdbc:mysql://localhost:3306/db_taxi_service";
+    private static final String DB_USER = "root";
+
+    static {
+        try {
+            Class.forName("com.mysql.cj.jdbc.Driver");
```
KostyaNikiforov, line 14: It can be constant also

KostyaNikiforov (changes requested):

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,114 @@
...
+            return getManufacturersFromResultSet(statement);
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get all manufacturers from DB ", e);
+        }
+    }
+
+    @Override
+    public Manufacturer update(Manufacturer manufacturer) {
+        String stringRequest = "UPDATE manufacturers SET name = ?, "
+                + "country = ? WHERE id = ? AND is_deleted = false;";
```
KostyaNikiforov, line 68: Fix it everywhere
```suggestion
        String query = "UPDATE manufacturers SET name = ?, "
                + "country = ? WHERE id = ? AND is_deleted = false";

```

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #501, opened 2023-01-23
first impl

leonid-topal:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,32 @@
...
+        List<Manufacturer> allManufacturers = manufacturerDao.getAll();
+        for (Manufacturer allManufacturer : allManufacturers) {
+            System.out.println(allManufacturer);
+        }
+        Manufacturer createManufacturer = manufacturerDao.create(manufacturer);
+        System.out.println(createManufacturer);
+        Optional<Manufacturer> getManufacturer = manufacturerDao.get(manufacturer.getId());
+        System.out.println("Get manufacturer by id: " + manufacturer.getId());
+        Manufacturer updateManufacturer = new Manufacturer("ZAZ", "Ukraine");
+        updateManufacturer.setId(manufacturer.getId());
```
yurii-dikalo, line 25: don't change id, change name or country instead. let id be autogenerated

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,114 @@
...
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class ManufacturerDaoImpl implements ManufacturerDao {
+    @Override
+    public List<Manufacturer> getAll() {
+        String querySelectAll = "SELECT * FROM manufacturers WHERE is_deleted = false";
+        List<Manufacturer> manufacturers = new ArrayList<>();
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getAllManufacturers =
```
yurii-dikalo, line 22:
```suggestion
                PreparedStatement statement =

```
and fix it everywhere

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,114 @@
...
+import mate.jdbc.exception.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class ManufacturerDaoImpl implements ManufacturerDao {
+    @Override
+    public List<Manufacturer> getAll() {
+        String querySelectAll = "SELECT * FROM manufacturers WHERE is_deleted = false";
```
yurii-dikalo, line 19:
```suggestion
        String query = "SELECT * FROM manufacturers WHERE is_deleted = false";

```
same for every other method

yurii-dikalo (changes requested):

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
KostyaNikiforov: Agree, we already see the context in the method signature

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
KostyaNikiforov: same

KostyaNikiforov (changes requested):
🔥

KostyaNikiforov (approved):

PR #520, opened 2023-02-16
jv-jdbc solution

nkharyshyn:
-created taxi_driver_db scheme with manufacturers table (columns id, name, country, is_deleted) and connected this DB to project
-created Manufacturer model with id, name and country fields 
-created ManufacturerDao interface and its implementation class with CRUD methods 
-used created @Dao annotation and Injector class to create an instance of ManufacturerDaoImpl class 
-created custom DataProcessingException that extends RuntimeException

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -1,7 +1,42 @@
 package mate.jdbc;
 
-public class Main {
-    public static void main(String[] args) {
+import java.util.List;
+import mate.jdbc.dao.ManufacturerDao;
+import mate.jdbc.lib.Injector;
+import mate.jdbc.models.Manufacturer;
 
+public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
+    public static void main(String[] args) {
+        ManufacturerDao manufacturerDao = (ManufacturerDao) injector
+                .getInstance(ManufacturerDao.class);
+        Manufacturer frenchManufacturer = new Manufacturer();
+        frenchManufacturer.setCountry("France");
+        frenchManufacturer.setName("Renault");
+
+        Manufacturer germanManufacturer = new Manufacturer();
+        germanManufacturer.setCountry("Germany");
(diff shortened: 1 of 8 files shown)


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,42 @@
...
+        }
+
+        manufacturerDao.get(2L).ifPresent(System.out::println);
+        manufacturerDao.get(10L).ifPresent(System.out::println);
+
+        manufacturerDao.getAll().forEach(System.out::println);
 
+        germanManufacturer.setName("Audi");
+        manufacturerDao.update(germanManufacturer);
+        manufacturerDao.delete(3L);
```
Ivan95kos, line 39: For the future, It is better not to use magic numbers as identifiers
```suggestion
        manufacturerDao.get(frenchManufacturer.getId()).ifPresent(System.out::println);
        manufacturerDao.get(germanManufacturer.getId()).ifPresent(System.out::println);

        manufacturerDao.getAll().forEach(System.out::println);

        germanManufacturer.setName("Audi");
        manufacturerDao.update(germanManufacturer);
        manufacturerDao.delete(italianManufacturer.getId());
```

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.exceptions.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.models.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class ManufacturerDaoImpl implements ManufacturerDao {
```
Ivan95kos, line 17: For the future, you need to create an impl package and move the implementation there

Ivan95kos (approved):
Great job

PR #528, opened 2023-02-22
Solution

kateryna-mykh:


src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,109 @@
...
+            updateStatement.executeUpdate();
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't update manufacturer " + manufacturer, e);
+        }
+        return manufacturer;
+    }
+
+    @Override
+    public boolean delete(Long id) {
+        String deleteRequest = "UPDATE manufacturers SET is_deleted = TRUE where id = ?;";
```
kozhukhovsky, line 88:
```suggestion
        String deleteRequest = "UPDATE manufacturers SET is_deleted = TRUE WHERE id = ?;";
```

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,109 @@
...
+        Manufacturer manufacturer = new Manufacturer();
+        try {
+            manufacturer.setName(resultSet.getString("name"));
+            manufacturer.setCountry(resultSet.getString("country"));
+            manufacturer.setId(resultSet.getObject("id", Long.class));
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't convert ResultSet to manufacturer ", e);
+        }
+        return manufacturer;
+    }
```
kozhukhovsky, line 108:
```suggestion
    private Manufacturer convertResultSet(ResultSet resultSet) throws SQLException {
        Manufacturer manufacturer = new Manufacturer();
        manufacturer.setName(resultSet.getString("name"));
        manufacturer.setCountry(resultSet.getString("country"));
        manufacturer.setId(resultSet.getObject("id", Long.class));
       
        return manufacturer;
    }
```

We have exception handling in each of our CRUD methods. This method is called everywhere in the try block, so we can throw the exception further and not handle it here.

kozhukhovsky (changes requested):

olekskov (approved):
Good job!

PR #531, opened 2023-02-26
made connection with DB, added model of Manufacturer, implmented DAO …

YevheniiKilovyi:
…for Manufacturer model and added costum DataProcessingException


olekskov (approved):
Good job!

okuzan (approved):
Good job!

PR #534, opened 2023-03-01
Create connection with DB, implement CRUD methods

katerynapoli:


src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,119 @@
...
+    @Override
+    public Optional<Manufacturer> get(Long id) {
+        String getRequest = "SELECT * FROM manufacturers WHERE is_deleted = false AND id = ?;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getManufacturerStatement = connection
+                        .prepareStatement(getRequest)) {
+            getManufacturerStatement.setLong(1, id);
+            ResultSet resultSet = getManufacturerStatement.executeQuery();
+            if (resultSet.next()) {
+                return Optional.of(getManufacturer(resultSet));
```
Ivan95kos, line 50: For the future, it is better to use the method Optional.ofNullable()
```suggestion
                return Optional.ofNullable(getManufacturer(resultSet));
```

Ivan95kos (approved):
Great job

PR #538, opened 2023-03-09
Added solution to jv-jdbc task

serhii-shyian:


src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,113 @@
...
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can`t insert new "
+                    + manufacturer + " to db ", e);
+        }
+    }
+
+    @Override
+    public Optional<Manufacturer> get(Long id) {
+        String request = "SELECT * FROM manufacturers WHERE id = ?;";
```
olekskov, line 40: we shouldn't get deleted objects even by id

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,113 @@
...
+    private Manufacturer getInstanceManufacturer(ResultSet resultSet) {
+        try {
+            return new Manufacturer(
+                    resultSet.getObject("id", Long.class),
+                    resultSet.getString("name"),
+                    resultSet.getString("country"));
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can`t get data from "
+                    + resultSet, e);
+        }
```
olekskov, line 111: we perform mentioned logic in place that already in try/catch block for SQLException of parent method so no need to repeat it here, we can just add throws to method header and it will be handled there

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,113 @@
...
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can`t get all manufacturers from db", e);
+        }
+        return allManufacturers;
+    }
+
+    @Override
+    public Manufacturer update(Manufacturer manufacturer) {
+        String query = "UPDATE manufacturers SET name = ?, country = ? WHERE id = ?;";
```
olekskov, line 73: we shouldn't allow update data on deleted entries

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,113 @@
...
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can`t update "
+                    + manufacturer + " from db ", e);
+        }
+        return manufacturer;
+    }
+
+    @Override
+    public boolean delete(Long id) {
+        String query = "UPDATE manufacturers SET is_deleted = true WHERE id = ?;";
```
olekskov, line 90: it's better to also add check for is_deleted flag here not to return true in case of repetitive delete of same object

olekskov (changes requested):

src/main/java/mate/jdbc/dao/ManufacturerDao.java
```diff
@@ -0,0 +1,19 @@
+package mate.jdbc.dao;
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Manufacturer;
+
+@Dao
+public interface ManufacturerDao {
+    void create(Manufacturer manufacturer);
```
uliana-andrushchuk, line 10: Please, return created manufacturer from this method

uliana-andrushchuk (changes requested):

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,34 @@
...
+
 public class Main {
+    private static final Injector injector =
+            Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
+        ManufacturerDao manufacturerDao
+                = (ManufacturerDao) injector.getInstance(ManufacturerDao.class);
+
+        Manufacturer carManufacturer = new Manufacturer();
```
okuzan, line 17:
```suggestion
        Manufacturer teslaManufacturer = new Manufacturer();
```

okuzan (approved):
Well done!

PR #539, opened 2023-03-09
Implements dao layer for manufacturer

GalubaGSV:


src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,114 @@
...
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class ManufacturerDaoImpl implements ManufacturerDao {
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        String insertManufacturerRequest = "INSERT INTO manufacturers(name, country)"
```
maxline, line 20: I would name it createManufacturerQuery . It is better to use `query` word instead `request`, `request` usually means http request

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @Override
+    public boolean delete(Long id) {
+        String deleteManufacturerRequest = "UPDATE manufacturers SET is_deleted = true "
+                + "WHERE is_deleted = false AND id = ?;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement createManufacturersStatement =
+                        connection.prepareStatement(deleteManufacturerRequest)) {
+            createManufacturersStatement.setLong(1, id);
+            return createManufacturersStatement.executeUpdate() >= 1;
```
maxline, line 67: not critical but using `>0` instead of `>=1` is more readable for me. the same in other places in the code

src/main/resources/init_db.sql
```diff
@@ -0,0 +1,9 @@
+CREATE DATABASE `taxi_services` /*!40100 DEFAULT CHARACTER SET utf8mb3 */ /*!80016 DEFAULT ENCRYPTION='N' */;
+
+CREATE TABLE `manufacturers` (
+  `id` bigint NOT NULL AUTO_INCREMENT,
+  `name` varchar(255) DEFAULT NULL,
+  `country` varchar(255) DEFAULT NULL,
+  `is_deleted` tinyint NOT NULL DEFAULT '0',
```
maxline, line 7: It can be used MySQL BOOLEAN data type here, which is the synonym of TINYINT(1). MYSQL automatically cast it to TINYINT. Please find more here https://www.mysqltutorial.org/mysql-boolean/

maxline (changes requested):
Good job! Let's improve the solution

petro-ushchuk (approved):
Good work 🔥

PR #540, opened 2023-03-09
Task solution was created.

mrGvvD:
Connection to db was established.
Manufacturer model was created.
ManufacturerDaoImpl was implemented.

### DIFF
--- a/src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
+++ b/src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
@@ -1,0 +1,118 @@
+package mate.jdbc.dao;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.exception.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class ManufacturerDaoImpl implements ManufacturersDao {
+    private static final int FIRST_INDEX = 1;
+    private static final int SECOND_INDEX = 2;
+    private static final int THIRD_INDEX = 3;
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        String query = "INSERT INTO manufacturers (name, country) VALUES (?, ?)";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement preparedStatement =
+                        connection.prepareStatement(query, Statement.RETURN_GENERATED_KEYS)) {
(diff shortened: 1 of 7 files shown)


src/main/java/mate/jdbc/dao/ManufacturersDao.java
```diff
@@ -0,0 +1,18 @@
+package mate.jdbc.dao;
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.model.Manufacturer;
+
+public interface ManufacturersDao {
```
olekskov, line 7: There is no reason to use plural form in this name

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,118 @@
...
+            return manufacturer;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Updating 'manufacturers' data was failed with id: "
+                    + manufacturer.getId(), e);
+        }
+    }
+
+    @Override
+    public boolean delete(Long id) {
+        String query = "UPDATE manufacturers SET is_deleted = true WHERE id = ?";
```
olekskov, line 93: Here also better to check is_deleted flag in order not to return true in case of repetitive delete on same object

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,118 @@
...
+
+    private Manufacturer getManufacturer(ResultSet resultSet) {
+        try {
+            String name = resultSet.getString("name");
+            String country = resultSet.getString("country");
+            Long id = resultSet.getObject(FIRST_INDEX, Long.class);
+            Manufacturer manufacturer = new Manufacturer(name, country);
+            manufacturer.setId(id);
+            manufacturer.setCountry(country);
+            manufacturer.setName(name);
```
olekskov, line 111: why are we setting name and country twice?

olekskov (changes requested):

src/main/java/mate/jdbc/dao/ManufacturerDao.java
```diff
@@ -0,0 +1,18 @@
+package mate.jdbc.dao;
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.model.Manufacturer;
+
+public interface ManufacturerDao {
+
```
Ivan95kos, line 8: For the future, remove the first empty line in all classes/interfaces/methods where it is present
```suggestion
```

src/main/java/mate/jdbc/exception/DataProcessingException.java
```diff
@@ -0,0 +1,8 @@
+package mate.jdbc.exception;
+
+public class DataProcessingException extends RuntimeException {
+    public DataProcessingException(String message, Exception e) {
+        super(message, e);
+    }
+
```
Ivan95kos, line 7: For the future, remove the last empty line in all classes/interfaces/methods where it is present
```suggestion
```

Ivan95kos (approved):

PR #551, opened 2023-03-20
First mission objectives accomplished

IlkoGood:


src/main/java/mate/jdbc/dao/ManufacturerDao.java
```diff
@@ -0,0 +1,19 @@
...
+
+    Optional<Manufacturer> get(Long id);
+
+    List<Manufacturer> getAll();
+
+    Manufacturer update(Manufacturer manufacturer);
+
+    boolean delete(Long id);
+
+    void clear();
```
Nikitos787, line 18: I'm not sure that we need this method clear...

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,132 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                 PreparedStatement preparedStatement =
+                         connection.prepareStatement("TRUNCATE manufacturers")) {
+            preparedStatement.execute();
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't perform truncation", e);
+        }
+    }
+
+    private static Optional<Manufacturer> getManufacturer(ResultSet resultSet) throws SQLException {
```
Nikitos787, line 123: Interesting) If I were you I would return here just the instance of Manufacturer. But maybe we can do it in your way)

src/main/java/mate/jdbc/dao/ManufacturerDao.java
IlkoGood: We don't, but I needed it)
It's more convenient to work with empty table with droped values of Autoincrement

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
IlkoGood: I'm not sure either, we will see)

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.Optional;
+import mate.jdbc.exception.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class ManufacturerDaoImpl implements ManufacturerDao {
+    private static final String FORMAT_REQUEST_INSERT =
+            "INSERT INTO manufacturers(name, country) values(?, ?);";
```
sarakhmen, line 19: Let's fix query syntax in all places:
```suggestion
            "INSERT INTO manufacturers (name, country) VALUES (?, ?)";
```

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,132 @@
...
+    private static final String FORMAT_REQUEST_GET =
+            "SELECT * FROM manufacturers WHERE is_deleted = false AND id = ?;";
```
sarakhmen, line 21:
```suggestion
            "SELECT * FROM manufacturers WHERE is_deleted = FALSE AND id = ?";
```

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,132 @@
...
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't add to DB: " + manufacturer, e);
+        }
+        return manufacturer;
+    }
+
+    @Override
+    public Optional<Manufacturer> get(Long id) {
+        Optional<Manufacturer> manufacturerOptional = Optional.empty();
```
sarakhmen, line 51: I'd recommend removing this line and creating/returning Optional.empty() only in case no manufacturers with requested id are found in db

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,132 @@
...
+        List<Manufacturer> manufacturers = new ArrayList<>();
+        try (Connection connection = ConnectionUtil.getConnection();
+                 PreparedStatement preparedStatement =
+                         connection.prepareStatement(FORMAT_REQUEST_GET_ALL)) {
+            ResultSet resultSet = preparedStatement.executeQuery();
+            while (resultSet.next()) {
+                manufacturers.add(new Manufacturer(
+                        resultSet.getLong(1),
+                        resultSet.getString("name"),
+                        resultSet.getString("country")));
```
sarakhmen, line 77: let's use your getManufacturer method here. Also, I'd recommend renaming the method to parseManufacturerFromResultSet or something like this

sarakhmen (changes requested):
Overall good job, but let's slightly improve your solution :)

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,114 @@
...
+    private static final String FORMAT_REQUEST_INSERT =
+            "INSERT INTO manufacturers(name, country) VALUES(?, ?);";
+    private static final String FORMAT_REQUEST_GET =
+            "SELECT * FROM manufacturers WHERE is_deleted = FALSE AND id = ?";
+    private static final String FORMAT_REQUEST_GET_ALL =
+            "SELECT * FROM manufacturers WHERE is_deleted = FALSE;";
+    private static final String FORMAT_REQUEST_UPDATE =
+            "UPDATE manufacturers SET name = ?, country = ? WHERE id = ? AND is_deleted = FALSE;";
+    private static final String FORMAT_REQUEST_DELETE_SOFT =
+            "UPDATE manufacturers SET is_deleted = TRUE WHERE id = ? AND is_deleted = FALSE;";
```
nacenik, line 27: Constants for your sql requests won't make your code easier to read, use local string in each method.[See here 13th bulletpoint](https://mate-academy.github.io/jv-program-common-mistakes/java-JDBC/jdbc-intro/JDBC-intro_checklist.html#common-mistakes-jv-jdbc)

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,114 @@
...
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class ManufacturerDaoImpl implements ManufacturerDao {
+    private static final String FORMAT_REQUEST_INSERT =
+            "INSERT INTO manufacturers(name, country) VALUES(?, ?);";
+    private static final String FORMAT_REQUEST_GET =
+            "SELECT * FROM manufacturers WHERE is_deleted = FALSE AND id = ?";
```
nacenik, line 21:
```suggestion
            "SELECT * FROM manufacturers WHERE is_deleted = FALSE AND id = ?;";
```

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,114 @@
...
+                 PreparedStatement preparedStatement =
+                         connection.prepareStatement(FORMAT_REQUEST_DELETE_SOFT)) {
+            preparedStatement.setLong(1, id);
+            return preparedStatement.executeUpdate() > 0;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't delete manufacturer by id: " + id, e);
+        }
+    }
+
+    private Manufacturer makeManufacturerFromInput(ResultSet resultSet) throws SQLException {
```
nacenik, line 108:
```suggestion
    private Manufacturer createManufacturerFromInput(ResultSet resultSet) throws SQLException {
```

nacenik (changes requested):
Good solution! Let’s improve it 😉

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
IlkoGood: Please make deal with this inside your cosy company)
See @sarakhmen review on same ";"

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
nacenik: The main point, that it is hard to read code, when you have queries as a constant. Just imagine, that you have not 5 methods, but 20. Will it be easy to check your queries and find possible mistake?

IlkoGood: It's much more convenient to tap on constant and read the line all at once.
Instead I need to create multiple lines of code with concatenation due to intendation rules, and this become the moment, when it is hard to read.
I will do as you say, but I really don't agree.

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,106 @@
...
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class ManufacturerDaoImpl implements ManufacturerDao {
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        try (Connection connection = ConnectionUtil.getConnection();
+                 PreparedStatement preparedStatement = connection.prepareStatement(
+                         "INSERT INTO manufacturers(name, country) VALUES(?, ?);",
```
nacenik, line 23: SQL query should be variable but inside the method. Fix it for all queries 
image

nacenik (changes requested):
Just fix some mistakes :)

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
IlkoGood: That was some kind of mind malfunction, I'll repair this)

nacenik (approved):
LGTM! 🚀

PR #556, opened 2023-03-22
Created DAO/impl, model, and exception

OleksiiKlymenko:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,27 @@
...
+                List.of(new Manufacturer("Toyota", "Japan"),
+                        new Manufacturer("Ford", "USA"),
+                        new Manufacturer("Accent", "South Korea"));
+        ManufacturerDao manufacturerDao
+                = (ManufacturerDao) injector.getInstance(ManufacturerDao.class);
 
+        for (Manufacturer manufacturer: listOfManufacturers) {
+            manufacturerDao.create(manufacturer);
+        }
+        System.out.println(manufacturerDao.get(1L));
```
Ivan95kos, line 22: It is a bad practice to use magical IDs
```suggestion
        System.out.println(manufacturerDao.get(manufacturer.getId()));
```

Ivan95kos (approved):

PR #558, opened 2023-03-22
Implemented DAO system for Manufacturer table

De1eF:
along with scheme and table itself

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,6 +18,14 @@
         </maven.checkstyle.plugin.configLocation>
     </properties>
 
+    <dependencies>
+        <dependency>
+            <groupId>com.mysql</groupId>
+            <artifactId>mysql-connector-j</artifactId>
+            <version>8.0.32</version>
+        </dependency>
+
+    </dependencies>
     <build>
         <plugins>
             <plugin>
--- a/src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
+++ b/src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
@@ -1,0 +1,117 @@
+package mate.jdbc.dao;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.DataProcessingException;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class ManufacturerDaoImpl implements ManufacturerDao {
+
+    @Override
+    public Manufacturer create(Manufacturer manufaturer) {
+        String createManufacturerRequest =
+                "INSERT INTO manufacturers(name, country, is_deleted) VALUES(?, ?, ?)";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement createManufacturer =
+                        connection.prepareStatement(createManufacturerRequest,
+                             Statement.RETURN_GENERATED_KEYS)) {
+            createManufacturer.setString(1, manufaturer.getName());
+            createManufacturer.setString(2, manufaturer.getCountry());
+            createManufacturer.setString(3, "0");
+            createManufacturer.executeUpdate();
+            ResultSet generatedKeys = createManufacturer.getGeneratedKeys();
+            if (generatedKeys.next()) {
+                Long id = generatedKeys.getObject(1, Long.class);
+                manufaturer.setId(id);
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't create manufacturer: " + manufaturer, e);
+        }
+        return manufaturer;
+    }
+
+    @Override
+    public Optional<Manufacturer> get(Long id) {
+        String getManufacturerRequest =
+                "SELECT * FROM manufacturers WHERE id = ? AND is_deleted = 0";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getManufacturer =
+                        connection.prepareStatement(getManufacturerRequest,
+                             Statement.RETURN_GENERATED_KEYS)) {
+            getManufacturer.setString(1, id.toString());
+            ResultSet resultSet = getManufacturer.executeQuery();
+            Optional<Manufacturer> returnedManufacturer = Optional.empty();
+            if (resultSet.next()) {
+                returnedManufacturer = Optional.of(
+                         new Manufacturer(resultSet.getObject("id", Long.class),
+                                 resultSet.getString("name"),
+                                 resultSet.getString("country")));
+            }
+            return returnedManufacturer;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't fetch manufacturer with id: " + id, e);
+        }
+    }
+
+    @Override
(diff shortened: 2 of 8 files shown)


src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,117 @@
...
+                             Statement.RETURN_GENERATED_KEYS)) {
+            getManufacturer.setString(1, id.toString());
+            ResultSet resultSet = getManufacturer.executeQuery();
+            Optional<Manufacturer> returnedManufacturer = Optional.empty();
+            if (resultSet.next()) {
+                returnedManufacturer = Optional.of(
+                         new Manufacturer(resultSet.getObject("id", Long.class),
+                                 resultSet.getString("name"),
+                                 resultSet.getString("country")));
+            }
```
petro-ushchuk, line 58: Let's keep code without duplications, review the methods and try to extract the same logic to separate methods.

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,117 @@
...
+    public Manufacturer create(Manufacturer manufaturer) {
+        String createManufacturerRequest =
+                "INSERT INTO manufacturers(name, country, is_deleted) VALUES(?, ?, ?)";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement createManufacturer =
+                        connection.prepareStatement(createManufacturerRequest,
+                             Statement.RETURN_GENERATED_KEYS)) {
+            createManufacturer.setString(1, manufaturer.getName());
+            createManufacturer.setString(2, manufaturer.getCountry());
+            createManufacturer.setString(3, "0");
```
petro-ushchuk, line 29: You should use `true` and `false` instead of `0` and `1`, check common mistakes

src/main/resources/init_db.sql
```diff
@@ -0,0 +1,9 @@
+CREATE DATABASE `db_taxi` /*!40100 DEFAULT CHARACTER SET utf8mb3 */ /*!80016 DEFAULT ENCRYPTION='N' */;
+
+CREATE TABLE `manufacturers` (
+  `id` int NOT NULL AUTO_INCREMENT,
+  `name` varchar(128) DEFAULT NULL,
+  `country` varchar(128) DEFAULT NULL,
+  `is_deleted` tinyint DEFAULT NULL,
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB AUTO_INCREMENT=17 DEFAULT CHARSET=utf8mb3;
```
petro-ushchuk, line 9: Bad idea to copy-paste init db from MySQL generator. What if the db would change?

petro-ushchuk (changes requested):
Good work! Very clean code, but we should improve these moments.

pom.xml
```diff
@@ -18,6 +18,14 @@
         </maven.checkstyle.plugin.configLocation>
     </properties>
 
+    <dependencies>
+        <dependency>
+            <groupId>com.mysql</groupId>
+            <artifactId>mysql-connector-j</artifactId>
+            <version>8.0.32</version>
+        </dependency>
+
```
nacenik, line 27: empty line

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    @Override
+    public Optional<Manufacturer> get(Long id) {
+        String getManufacturerRequest =
+                "SELECT * FROM manufacturers WHERE id = ? AND is_deleted = FALSE";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getManufacturer =
+                        connection.prepareStatement(getManufacturerRequest,
+                             Statement.RETURN_GENERATED_KEYS)) {
```
nacenik, line 49: image

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    @Override
+    public boolean delete(Long id) {
+        String deleteManufacturerRequest =
+                "UPDATE manufacturers SET is_deleted = TRUE WHERE id = ? AND is_deleted = FALSE";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement deleteManufacturer =
+                        connection.prepareStatement(deleteManufacturerRequest,
+                             Statement.RETURN_GENERATED_KEYS)) {
```
nacenik, line 107: delete `Statement.RETURN_GENERATED_KEYS`

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,125 @@
...
+            deleteManufacturer.setString(1, id.toString());
```
nacenik, line 108: You should use `setLong()` instead of `setString()`

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,125 @@
...
+
+    private Manufacturer getManufacturerFromResultSet(ResultSet resultSet) {
+        try {
+            return new Manufacturer(resultSet.getObject("id", Long.class),
+                    resultSet.getString("name"),
+                    resultSet.getString("country"));
+        } catch (SQLException e) {
+            throw new DataProcessingException(
+                    "Can't create a new Manufacturer object from: " + resultSet, e);
+        }
```
nacenik, line 123: You don't need to have `try-catch` block here, cause all code, where you call this method, contains can catch SQLException. 
Just add it to the method signature

src/main/java/mate/jdbc/model/Manufacturer.java
```diff
@@ -0,0 +1,51 @@
...
+    public Manufacturer(String name, String country) {
+        this.name = name;
+        this.country = country;
+    }
+
+    public Manufacturer(Long id, String name, String country) {
+        this.id = id;
+        this.name = name;
+        this.country = country;
+    }
```
nacenik, line 17: reuse constructor 
```suggestion
    public Manufacturer(Long id, String name, String country) {
        this(name, country);
        this.id = id;
    }
```

src/main/resources/init_db.sql
```diff
@@ -0,0 +1,9 @@
+CREATE DATABASE `db_taxi` /*!40100 DEFAULT CHARACTER SET utf8mb3 */ /*!80016 DEFAULT ENCRYPTION='N' */;
+
+CREATE TABLE `manufacturers` (
+  `id` int NOT NULL AUTO_INCREMENT,
+  `name` varchar(128) DEFAULT NULL,
+  `country` varchar(128) DEFAULT NULL,
+  `is_deleted` tinyint DEFAULT NULL,
```
nacenik, line 7: SQL variable types should be uppercase

nacenik (changes requested):
Let’s improve your solution ;)

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
nacenik: Not fixed. `Statement.RETURN_GENERATED_KEYS` should be only in create method

nacenik (approved):
Great job! You can improve your solution if you want 😉

PR #578, opened 2023-03-23
impl jdbc with basic crud operations

EliSoll:


src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,125 @@
...
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get all manufacturers from taxi_DB", e);
+        }
+        return allManufacturers;
+    }
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        String insertFormatsRequest = "INSERT INTO manufacturers(name, country)"
+                + " values(?, ?)";
```
redmanO-o, line 62: Think that values may be in UpperCase

redmanO-o (approved):
looks like a nice job ;)

pom.xml
```diff
@@ -17,7 +17,13 @@
             https://raw.githubusercontent.com/mate-academy/style-guides/master/java/checkstyle.xml
         </maven.checkstyle.plugin.configLocation>
     </properties>
-
+    <dependencies>
+    <dependency>
+        <groupId>mysql</groupId>
+        <artifactId>mysql-connector-java</artifactId>
+        <version>8.0.32</version>
+    </dependency>
```
Andriy-dz, line 25:
```suggestion
       <dependency>
           <groupId>mysql</groupId>
           <artifactId>mysql-connector-java</artifactId>
           <version>8.0.32</version>
       </dependency>
```

Andriy-dz: it will be nicer

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    private Manufacturer retrieveManufacturer(ResultSet set) throws SQLException {
+        String name = set.getString(COLUMN_NAME);
+        String country = set.getString(COLUMN_COUNTRY);
+        Long id = set.getObject(COLUMN_ID, Long.class);
+        return new Manufacturer(id, name, country);
+    }
+
+    private static void setStrings(Manufacturer manufacturer, PreparedStatement
```
Andriy-dz, line 120: it's a good idea to create a separate method. But why static? and I would recommend calling parseNameAndCounty

Andriy-dz (changes requested):

PR #588, opened 2023-03-23
Dao created and implemented

AnonimToster:


src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,109 @@
...
+    private static final String CREATE_QUERY =
+            "INSERT INTO manufacturer(name, country) VALUES(?, ?);";
+    private static final String GET_QUERY =
+            "SELECT * FROM manufacturer WHERE id = ? AND NOT is_deleted;";
+    private static final String GET_ALL_QUERY =
+            "SELECT * FROM manufacturer WHERE NOT is_deleted";
+    private static final String UPDATE_QUERY =
+            "UPDATE manufacturer SET name = ?, country = ? WHERE id = ?;";
+    private static final String DELETE_QUERY =
+            "UPDATE manufacturer SET is_deleted = true WHERE id = ?;";
```
AleksandrLts, line 27: I think it is better to create these variables directly in the methods where they are used, since they are only used once

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement preparedStatement =
```
AleksandrLts, line 32: variables should be informative

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,109 @@
...
+                String name = resultSet.getString("name");
+                String country = resultSet.getString("country");
+                return Optional.of(new Manufacturer(id, name, country));
+            } else {
+                return Optional.empty();
+            }
+        } catch (SQLException e) {
+            throw new RuntimeException("Can't get manufacturer with id " + id, e);
+        }
+    }
```
AleksandrLts, line 65:
```suggestion
            }
        } catch (SQLException e) {
            throw new RuntimeException("Can't get manufacturer with id " + id, e);
        }
        return Optional.empty();
    }
```
Looks better

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Override
+    public List<Manufacturer> getAll() {
+        try (Connection connection = ConnectionUtil.getConnection();
+                Statement statement = connection.createStatement()) {
+            ResultSet resultSet = statement.executeQuery(GET_ALL_QUERY);
+            List<Manufacturer> manufacturers = new ArrayList<>();
```
AleksandrLts, line 72: it is redundant line in try block, move it outside the block try with resources

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,109 @@
...
+            ResultSet resultSet = preparedStatement.executeQuery();
+            if (resultSet.next()) {
+                String name = resultSet.getString("name");
+                String country = resultSet.getString("country");
+                return Optional.of(new Manufacturer(id, name, country));
+            } else {
+                return Optional.empty();
+            }
+        } catch (SQLException e) {
+            throw new RuntimeException("Can't get manufacturer with id " + id, e);
```
AleksandrLts, line 63: you need to create your custom exception, check the task conditions

AleksandrLts (changes requested):
Good job, fix a bit!)

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
AnonimToster: but why? If execute will throw exception, creating redundant list will be useless and just waste resources. And in this case List created closer to place where it used.

AleksandrLts (approved):

src/main/java/mate/jdbc/dao/ManufacturerDao.java
```diff
@@ -0,0 +1,19 @@
+package mate.jdbc.dao;
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Manufacturer;
+
+@Dao
+public interface ManufacturerDao {
```
KseniiaMakarova, line 9: should we use annotation with interface or only with implementation?

src/main/resources/init_db.sql
```diff
@@ -0,0 +1,9 @@
+CREATE DATABASE `cars_db` /*!40100 DEFAULT CHARACTER SET utf8mb3 COLLATE utf8mb3_bin */ /*!80016 DEFAULT ENCRYPTION='N' */;
+
+CREATE TABLE `manufacturer` (
+  `id` int NOT NULL AUTO_INCREMENT,
```
KseniiaMakarova, line 4: use bigint (this is long)

src/main/resources/init_db.sql
```diff
@@ -0,0 +1,9 @@
+CREATE DATABASE `cars_db` /*!40100 DEFAULT CHARACTER SET utf8mb3 COLLATE utf8mb3_bin */ /*!80016 DEFAULT ENCRYPTION='N' */;
+
+CREATE TABLE `manufacturer` (
```
KseniiaMakarova, line 3:
```suggestion
CREATE TABLE `manufacturers` (
```

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,106 @@
...
+            }
+        } catch (SQLException e) {
+            throw new DataQueryException("Can't create a " + manufacturer, e);
+        }
+        return manufacturer;
+    }
+
+    @Override
+    public Optional<Manufacturer> get(Long id) {
+        String getQuery = "SELECT * FROM manufacturer WHERE id = ? AND NOT is_deleted;";
```
KseniiaMakarova, line 42:
```suggestion
        String getQuery = "SELECT * FROM manufacturers WHERE id = ? AND is_deleted = FALSE;";
```
does it work with NOT is_deleted?

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,106 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getManufacturerStatement
+                        = connection.prepareStatement(getQuery)) {
+            getManufacturerStatement.setLong(1, id);
+            ResultSet resultSet = getManufacturerStatement.executeQuery();
+            if (resultSet.next()) {
+                String name = resultSet.getString("name");
+                String country = resultSet.getString("country");
+                return Optional.of(new Manufacturer(id, name, country));
```
KseniiaMakarova, line 51: move to private method

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,106 @@
...
+        String getAllQuery = "SELECT * FROM manufacturer WHERE NOT is_deleted";
+        try (Connection connection = ConnectionUtil.getConnection();
+                Statement getAllManufacturersStatement = connection.createStatement()) {
+            ResultSet resultSet = getAllManufacturersStatement.executeQuery(getAllQuery);
+            List<Manufacturer> manufacturers = new ArrayList<>();
+            while (resultSet.next()) {
+                Long id = resultSet.getObject("id", Long.class);
+                String name = resultSet.getString("name");
+                String country = resultSet.getString("country");
+                manufacturers.add(new Manufacturer(id, name, country));
```
KseniiaMakarova, line 70: use your private method to form Manufacturer object

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,106 @@
...
+            }
+            return manufacturers;
+        } catch (SQLException e) {
+            throw new DataQueryException("Can't get manufacturers from DB", e);
+        }
+    }
+
+    @Override
+    public Manufacturer update(Manufacturer manufacturer) {
+        String update = "UPDATE manufacturer SET name = ?, country = ? WHERE id = ?;";
```
KseniiaMakarova, line 80:
```suggestion
        String update = "UPDATE manufacturers SET name = ?, country = ? WHERE id = ? AND is_deleted = FALSE;";
```

KseniiaMakarova (changes requested):

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
AnonimToster: yes, it work with NOT

src/main/resources/init_db.sql
AnonimToster: I don't understand why. I can change it, but every task from SQL basic had tables like "pupil", "dish", "book". I just don't understand this moment. But okay, I will just change it.

AnonimToster: SQL style guide advise to use single form too... It's confuses me
https://www.sqlstyle.guide/#tables

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
AnonimToster: I should use setters to form Manufacturer?

KseniiaMakarova (approved):

PR #590, opened 2023-03-23
implemented DAO for Manufacterer

SanykVolodymyr:


sarakhmen (changes requested):
Overall good job, but let's slightly improve your solution :)

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,45 @@
...
 
+import java.util.List;
+import mate.jdbc.dao.ManufacturerDao;
+import mate.jdbc.lib.Injector;
+import mate.jdbc.models.Manufacturer;
+
 public class Main {
+    private static Manufacturer bmwManufacturer = new Manufacturer("BMW","Germany");
+    private static Manufacturer volvoManufacturer = new Manufacturer("Volvo","Sweden");
+    private static Manufacturer fordManufacturer = new Manufacturer("Ford","USA");
```
KseniiaMakarova, line 11: it is better to make these 3 variables local

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,45 @@
...
+        );
+
+        System.out.println("Test create:");
+        for (Manufacturer manufacturer : manufacturers) {
+            System.out.println("Before: " + manufacturer);
+            manufacturer = manufacturerDao.create(manufacturer);
+            System.out.println("After:  " + manufacturer);
+        }
+
+        Long id = 34L;
```
KseniiaMakarova, line 28:
```suggestion
        Long id = bmwManufacturer.getId();
```
please don't hardcode id
https://mateacademy.slack.com/archives/C04L1TMSAKD/p1679918103022809

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,45 @@
...
+        System.out.println("return = " + manufacturerDao.get(id));
+
+        System.out.println("\nTest delete manufacturer by id = " + id);
+        System.out.println("return = " + manufacturerDao.delete(id));
+        System.out.println("after get(" + id + ") = " + manufacturerDao.get(id));
+
+        Manufacturer manufacturerUpdate
+                = new Manufacturer(35L, "Seat", "Spain");
+        System.out.println("\nTest update manufacturer " + manufacturerUpdate);
+        System.out.println("return = " + manufacturerDao.update(manufacturerUpdate));
```
KseniiaMakarova, line 39:
```suggestion
        volvoManufacturer.setCountry("Spain");
        System.out.println("\nTest update manufacturer " + volvoManufacturer);
        System.out.println("return = " + manufacturerDao.update(volvoManufacturer));
```

we should update existing objects, and we should create new manufacturers using create()
https://mateacademy.slack.com/archives/C04L1TMSAKD/p1679918491713599

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,123 @@
...
+            updateManufacturerStatement.setString(2, manufacturer.getCountry());
+            updateManufacturerStatement.setLong(3, manufacturer.getId());
+            if (updateManufacturerStatement.executeUpdate() > 0) {
+                return manufacturer;
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't update manufacturer "
+                    + manufacturer + " to DB", e);
+        }
+        return null;
```
KseniiaMakarova, line 99: let's never return null
either throw Exception, or return manufacturer

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #605, opened 2023-04-23
added proper solution for this task

polikarpov13:


src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,135 @@
...
+    public boolean delete(Long id) {
+        String deleteFormatRequest =
+                "UPDATE manufacturers SET is_deleted = true WHERE id = ?;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement deleteFormatStatement = connection
+                        .prepareStatement(deleteFormatRequest, Statement.RETURN_GENERATED_KEYS)) {
+            deleteFormatStatement.setLong(1, id);
+            return deleteFormatStatement.executeUpdate() >= 1;
+        } catch (SQLException e) {
+            throw new RuntimeException("Can not delete format from DB", e);
```
Ivan95kos, line 132: Change the exception thrown to DataProcessingException instead of RuntimeException to maintain consistency.

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,25 @@
...
+
+        manufacturer.setId(1L);
+        manufacturer.setName("ALALA");
+        manufacturer.setCountry("ALAa");
+
+        ManufacturerDao manufacturerDao = (ManufacturerDao) injector
+                .getInstance(ManufacturerDao.class);
 
+        manufacturerDao.getAll().forEach(System.out::println);
+        //manufacturerDao.create(manufacturer);
```
Ivan95kos, line 21: It's better to avoid commented-out code in your final version. If you want to keep these lines for testing purposes, you can create unit tests to cover these cases.

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,25 @@
...
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Manufacturer;
+
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
+        Manufacturer manufacturer = new Manufacturer();
+
+        manufacturer.setId(1L);
```
Ivan95kos, line 13: It's not a good practice to set the `id` manually, as it should be auto-generated by the database. Please remove this line.

Ivan95kos (approved):

PR #615, opened 2023-05-08
HW 02 solution

IhorPetrytskyi333:


src/main/java/mate/jdbc/exception/DataProcessingException.java
```diff
@@ -0,0 +1,7 @@
+package mate.jdbc.exception;
+
+public class DataProcessingException extends RuntimeException {
+    public DataProcessingException(String message, Throwable ex) {
+        super(message,ex);
```
uliana-andrushchuk, line 5:
```suggestion
        super(message, ex);
```

uliana-andrushchuk (changes requested):

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,121 @@
...
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionsUtil;
+
+@Dao
+public class ManufacturerDaoImpl implements ManufacturerDao {
+    @Override
+
+    public Manufacturer create(Manufacturer manufacturer) {
+        String selectQuery = "SELECT id FROM manufacturers WHERE name = ? AND country = ?";
+        String insertQuery = "INSERT INTO manufacturers (name, country) VALUES (?, ?)";
```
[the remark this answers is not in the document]
olekskov: Not fixed. We shouldn't do manual checks for presence in DB. It should be done on DB side with proper constraints if required. 
For the moment you shouldn't worry about duplicates. Just leave saving logic and remove select part. In future you will be able to tune DB to reject attempt to save duplicates

src/main/java/mate/jdbc/exception/DataProcessingException.java
olekskov: not fixed

olekskov (changes requested):

kozhukhovsky (approved):

PR #619, opened 2023-05-16
jv-jdbc hw first solution

kcasPilot:


src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,133 @@
...
+            throw new DataProcessingException("The manufacturer can't be null");
+        }
+        if (manufacturer.getId() == null) {
+            throw new DataProcessingException(
+                    "Can't update data for manufacturer with id=NULL: " + manufacturer);
+        }
+        if (get(manufacturer.getId()).isEmpty()) {
+            throw new DataProcessingException(
+                    "The manufacturer: " + manufacturer + " doesn't exist in DB");
+        }
```
kozhukhovsky, line 89:
```suggestion
```

kozhukhovsky: Such checks must be done in the service

kozhukhovsky (approved):

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
kcasPilot: Ok, thanks

PR #625, opened 2023-05-25
task solved

MykytaKat:


src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,121 @@
...
+            throw new DataProcessingException("Can't get manufacturer by id " + id,e);
+        }
+    }
+
+    @Override
+    public List<Manufacturer> getAll() {
+        List<Manufacturer> manufacturers = new ArrayList<>();
+        try (Connection connection = ConnectionUnit.getConnection();
+                PreparedStatement getAllManufacturersStatement
+                        = connection.prepareStatement("SELECT * FROM manufacturers")) {
```
OlehPryshchepa, line 69: this query will return all information from table even deleted records, create separate query for example "getAllQuery", and add condition where is_deleted = false

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,121 @@
...
+
+    @Override
+    public boolean delete(Long id) {
+        String query = "UPDATE manufacturers SET is_deleted = TRUE WHERE id = ?";
+        try (Connection connection = ConnectionUnit.getConnection()) {
+            PreparedStatement preparedStatement = connection.prepareStatement(query);
+            preparedStatement.setLong(1, id);
+            int result = preparedStatement.executeUpdate();
+            preparedStatement.close();
+            return result != 0;
```
OlehPryshchepa, line 104: you can simplify to return preparedStatement.executeUpdate > 0

src/main/java/mate/jdbc/unit/ConnectionUnit.java
```diff
@@ -0,0 +1,28 @@
...
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.util.Properties;
+
+public class ConnectionUnit {
+    static {
+        try {
+            Class.forName("com.mysql.cj.jdbc.Driver");
```
OlehPryshchepa, line 11: please move all hardcoded connection parameters (included "user" and "password") in this class to the constants.

src/main/resources/init_db.sql
```diff
@@ -0,0 +1,10 @@
+CREATE DATABASE `db_taxi_service` /*!40100 DEFAULT CHARACTER SET utf8mb3 */ /*!80016 DEFAULT ENCRYPTION='N' */;
+
+CREATE TABLE `manufacturers` (
+                                 `id` bigint NOT NULL AUTO_INCREMENT,
```
OlehPryshchepa, line 4: please fix all sql keywords to uppercase below, e.g.:

OlehPryshchepa (changes requested):
Good job! let`s improve

OlehPryshchepa (approved):
Great job! keep pushing

sarakhmen (changes requested):
Good job, but let's slightly improve your solution :)

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,119 @@
...
+        }
+    }
+
+    private Manufacturer createManufacturer(ResultSet resultSet) {
+        Manufacturer manufacturer = new Manufacturer();
+        try {
+            manufacturer.setCountry(resultSet.getString("country"));
+            manufacturer.setId(resultSet.getObject("id", Long.class));
+            manufacturer.setName(resultSet.getString("name"));
+        } catch (SQLException e) {
```
maxline, line 114: Let's better to throw exception in method declaration line and get rid of `try catch` block inside of this method.
`private Manufacturer createManufacturer(ResultSet resultSet) throws SQLException {`
We already catch `SQLException` in the methods that call `createManufacturer()` method.

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,26 @@
...
+
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
+        ManufacturerDao manufacturerDao
+                = (ManufacturerDao) injector.getInstance(ManufacturerDao.class);
+        Manufacturer manufacturer = new Manufacturer();
+        manufacturer.setCountry("Czech");
+        manufacturer.setName("Skoda");
```
maxline, line 15: please create a constructor with two parameters and call it here instead of calling default constructor with setters.

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,26 @@
...
+import mate.jdbc.dao.ManufacturerDao;
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Manufacturer;
+
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
+        ManufacturerDao manufacturerDao
+                = (ManufacturerDao) injector.getInstance(ManufacturerDao.class);
```
maxline, line 12: You can add ab empty line after this ling to get code code more readable with separating blocks and grouping lines with some common logic.

maxline (changes requested):
Good job! Left just a few minor suggestions.

maxline (approved):
Well done! Approved

PR #627, opened 2023-05-27
task was implemented

MaxSkory:


src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,109 @@
...
+        String sqlRequest = "SELECT * FROM manufacturers WHERE is_deleted = FALSE;";
+        List<Manufacturer> manufacturers = new ArrayList<>();
+        try (Connection connection = ConnectionUtil.createConnection();
+                 PreparedStatement statement = connection.prepareStatement(sqlRequest)) {
+            ResultSet resultSet = statement.executeQuery();
+            while (resultSet.next()) {
+                Long id = resultSet.getObject("id", Long.class);
+                String name = resultSet.getString("name");
+                String country = resultSet.getString("country");
+                manufacturers.add(new Manufacturer().setId(id).setName(name).setCountry(country));
```
olekskov, line 32: This logic is nearly identical to lines 47-49.
Please introduce separate private method that will convert resultSet to model and use it in both get() and getAll() methods

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,109 @@
...
+                return null;
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't update record for " + manufacturer, e);
+        }
+        return manufacturer;
+    }
+
+    public boolean delete(Long id) {
+        String sqlRequest = "UPDATE manufacturers SET is_deleted = TRUE WHERE id = ?;";
```
olekskov, line 95: Would be great to add check for is_deleted here also to receive proper result (false) in case user tries to delete same manufacturer multiple times

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,109 @@
...
+                manufacturer.setId(generatedKeys.getObject(1, Long.class));
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't create new record for " + manufacturer, e);
+        }
+        return manufacturer;
+    }
+
+    public Manufacturer update(Manufacturer manufacturer) {
+        String sqlRequest = "UPDATE manufacturers SET name = ?, country = ? WHERE id = ?;";
```
olekskov, line 78: We shouldn't update deleted records

olekskov (changes requested):

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,119 @@
...
+                manufacturers.add(getManufacturer(resultSet));
+            }
+            return manufacturers;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't retrieve records from DB", e);
+        }
+    }
+
+    public Optional<Manufacturer> get(Long id) {
+        String sqlRequest = "SELECT * FROM manufacturers WHERE id = ? AND is_deleted = FALSE;";
```
okuzan, line 38:
```suggestion
        String query = "SELECT * FROM manufacturers WHERE id = ? AND is_deleted = FALSE;";
```
when we talk about DB we call it query rather than request

src/main/resources/init_db.sql
```diff
@@ -0,0 +1,23 @@
...
+) ENGINE=InnoDB AUTO_INCREMENT=44 DEFAULT CHARSET=UTF8MB4;
+CREATE TABLE IF NOT EXISTS taxi_service.drivers (
+                           `id` bigint NOT NULL AUTO_INCREMENT,
+                           `first_name` varchar(255) DEFAULT NULL,
+                           `last_name` varchar(255) DEFAULT NULL,
+                           `car` bigint DEFAULT NULL,
+                           PRIMARY KEY (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=UTF8MB4;
+
+
```
okuzan, line 23: extra lines

okuzan (approved):
Good job!

PR #634, opened 2023-06-04
All task done

vrnvv:


src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,117 @@
...
+    @Override
+    public Manufacturer update(Manufacturer manufacturer) {
+        String updateRequest
+                = "UPDATE manufacturers SET name = ?, "
+                + "country = ? WHERE is_deleted = FALSE AND id = ?";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement updateManufacturerStatement
+                        = connection.prepareStatement(updateRequest)) {
+            updateManufacturerStatement.setString(1, manufacturer.getName());
+            updateManufacturerStatement.setString(2, manufacturer.getName());
```
olekskov, line 85: In 'updateManufacturerStatement.setString(2, manufacturer.getName())', 'manufacturer.getName()' should be changed to 'manufacturer.getCountry()'.

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,117 @@
...
+            updateManufacturerStatement.setLong(3, manufacturer.getId());
+            return updateManufacturerStatement.executeUpdate() > 0 ? manufacturer : null;
```
olekskov, line 87: It's not best practice to return null. If you wish to explicitly indicate that record for update doesn't exist then throw DataProcessingException with relevant message

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,117 @@
...
+
+    @Override
+    public boolean delete(Long id) {
+        String deleteRequest
+                = "UPDATE manufacturers SET is_deleted = TRUE WHERE id = ?";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement deleteManufacturersStatement
+                        = connection.prepareStatement(deleteRequest)) {
+            deleteManufacturersStatement.setLong(1, id);
+            return deleteManufacturersStatement.executeUpdate() >= 0;
```
olekskov, line 102: In 'return deleteManufacturersStatement.executeUpdate() >= 0;', the condition should be '> 0' instead of '>= 0' to make sure the method returns true only if at least one record is affected.

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,117 @@
...
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't create new record for " + manufacturer, e);
+        }
+        return manufacturer;
+    }
+
+    @Override
+    public Optional<Manufacturer> get(Long id) {
+        String insertRequest
```
olekskov, line 42: Variable name 'insertRequest' should be changed to 'getRequest' for better readability.

olekskov (changes requested):

Rommelua (approved):

PR #641, opened 2023-06-09
add jdbc task solution

AnastasiiaVariuchuk:


src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,118 @@
...
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get manufacturer from DB by id: " + id, e);
+        }
+    }
+
+    @Override
+    public List<Manufacturer> getAll() {
+        List<Manufacturer> allManufacturers = new ArrayList<>();
+        try (Connection connection = ConnectionUtil.getConnection();
+                 PreparedStatement statement
```
ArinaPolna, line 71: your solution is great) just one minor comment: according to the video lesson when we dont have dependency from our input parameters we can use just a simple Statement class to create our statement so as an option you can write something like this: Statement statement = connection.createStatement()) {
            ResultSet resultSet = statement.executeQuery(getALLManufacturersRequest). But i think it`s not critical in this situation. And also it`s a good practise to give us statements some specific name. You can use here getAllManufacturersStatement for example and also rename your statements in other method`s

ArinaPolna (approved):

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,26 @@
...
+        manufacturerDao.create(volkswagen);
+        manufacturerDao.create(kia);
+        manufacturerDao.create(toyota);
+        manufacturerDao.create(bmw);
+        manufacturerDao.getAll().stream().forEach(System.out::println);
+        manufacturerDao.delete(bmw.getId());
+        manufacturerDao.getAll().stream().forEach(System.out::println);
     }
 }
+
```
sarakhmen, line 26:
```suggestion
```

sarakhmen (approved):
Excellent job ;)

PR #642, opened 2023-06-09
create connection to database

ArinaPolna:


src/main/java/mate/jdbc/dao/ManufacturerDaoImp.java
```diff
@@ -0,0 +1,117 @@
...
+    public Optional<Manufacturer> get(Long id) {
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getManufacturer = connection.prepareStatement(
+                          getManufacturerRequest)) {
+            getManufacturer.setLong(1, id);
+            ResultSet resultSet = getManufacturer.executeQuery();
+            if (resultSet.next()) {
+                String name = resultSet.getString("name");
+                String country = resultSet.getString("country");
+                Manufacturer manufacturer = new Manufacturer(id, name, country);
```
AnastasiiaVariuchuk, line 59: in the method get and getAll you use a construction 
Manufacturer manufacturer = new Manufacturer(id, name, country) 
to create a manufactory, maybe it makes sense to separate it into a method
private Manufacturer createManufacturer(ResultSet resultSet)
but its just my recommendation

src/main/java/mate/jdbc/Main.java
```diff
@@ -5,3 +5,4 @@ public static void main(String[] args) {
```
AnastasiiaVariuchuk, line 5: In my opinion, it is worth showing the work of some methods here

AnastasiiaVariuchuk (commented):
Your solution is absolutely good, just minor comments)

src/main/java/mate/jdbc/dao/ManufacturerDaoImp.java
ArinaPolna: good idea) i wiil implement it)

AnastasiiaVariuchuk (approved):

sarakhmen (changes requested):
Let's improve your solution :)

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,27 @@
...
+    public static void main(String[] args) {
+        ManufacturerDao manufacturerDao = (ManufacturerDao)
+                injector.getInstance(ManufacturerDao.class);
+        manufacturerDao.create(new Manufacturer("audi", "Italy"));
+        manufacturerDao.create(new Manufacturer("volvo", "Germany"));
+        manufacturerDao.delete(2L);
+        manufacturerDao.update(new Manufacturer(7L, "shkoda", "Japan"));
+        Optional<Manufacturer> manufacturerWithIndex9 = manufacturerDao.get(9L);
+        Optional<Manufacturer> manufacturerToCompare = Optional.ofNullable(
+                new Manufacturer(9L, "volvo", "Germany"));
```
sarakhmen, line 21: here in your example/test code let's not use hardcoded int literals as id parameter. I'd recommend creating and assigning the manufacturer to some variable first. And then you can retrieve from this variable an id field and pass it to the corresponding methods

sarakhmen (approved):
Good job, you can improve your solution if you want :)

PR #643, opened 2023-06-09
Created a connection to Database, implemented a ManufacturerDao

viprko:


ViktorBykh (changes requested):
Everything looks well)
I have a few suggestions

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,116 @@
...
+
+    @Override
+    public Manufacturer update(Manufacturer manufacturer) {
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement updateManufacturerStatement = connection.prepareStatement(
+                        UPDATE_REQUEST)) {
+            updateManufacturerStatement.setString(1, manufacturer.getName());
+            updateManufacturerStatement.setString(2, manufacturer.getCountry());
+            updateManufacturerStatement.setLong(3, manufacturer.getId());
+            updateManufacturerStatement.executeUpdate();
```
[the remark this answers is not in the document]
viprko: I can't remove it, cos this is the main line that launches requests, without it, we would not change the DB. Also, I'm not sure about "checking how many rows we changed", cos it seems like it is easier to hack. Let's imagine it's accounts DB, I try to update the password, and the program throws an exception about 0 rows being changed, so now I know there is no such account, and make a script to identity what accounts exist.

ViktorBykh (approved):
Everything looks great)

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,39 @@
...
+        Manufacturer newManufacturer = new Manufacturer(5L, "Land Rover", "United Kingdom");
 
+        Optional<Manufacturer> getManufacturerById = manufacturerDao.get(5L);
+        System.out.println(getManufacturerById);
+
+        manufacturerDao.update(newManufacturer);
+        System.out.println(getManufacturerById);
+
+        boolean isDeleted = manufacturerDao.delete(8L);
+        System.out.println(isDeleted);
```
maxline, line 36: Let's also print the deleted with `manufacturerDao.get(id)`, or print all elements to see the on of them was deleted.

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,119 @@
...
+import mate.jdbc.exceptions.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class ManufacturerDaoImpl implements ManufacturerDao {
+    private static final String INSERT_REQUEST =
+            "INSERT INTO manufacturers(name, country) VALUES(?, ?)";
+    private static final String GET_SINGLE_MANUFACTURER_REQUEST =
```
maxline, line 19:
```suggestion
    private static final String GET_BY_ID_REQUEST =
```

maxline (changes requested):
Good job! Let's improve the solution a bit

sarakhmen (approved):
Good job :)

PR #644, opened 2023-06-09
done task

ViktoriiaSemenchenko:


src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,108 @@
...
+        String createQuery = "INSERT INTO manufacturers(name, country) VALUES (?, ?);";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement createStatement =
+                        connection.prepareStatement(createQuery, Statement.RETURN_GENERATED_KEYS)) {
+            createStatement.setString(1, manufacturer.getName());
+            createStatement.setString(2, manufacturer.getCountry());
+            createStatement.executeUpdate();
+            ResultSet generatedKeys = createStatement.getGeneratedKeys();
+            if (generatedKeys.next()) {
+                Long id = generatedKeys.getObject(1, Long.class);
```
Gamemod13, line 30: better use name of column than index in ResultSet

src/main/java/mate/jdbc/util/ConnectionUtil.java
```diff
@@ -0,0 +1,29 @@
...
+        } catch (ClassNotFoundException e) {
+            throw new RuntimeException("Can`t load JDBC driver to MySql", e);
+        }
+    }
+
+    public static Connection getConnection() {
+        try {
+            Properties dbProperties = new Properties();
+            dbProperties.put("user", "root");
+            dbProperties.put("password", "academy23");
```
Gamemod13, line 22: you can replace user-password in constant fields аor easier replacement if necessary

Gamemod13 (changes requested):
Good work, but let's slighty improve your solution)

Gamemod13 (approved):
Good, you can use index for your ResultSet I think

sarakhmen (approved):
Excellent job, you can improve your solution if you want :)

PR #648, opened 2023-06-09
Solved solution with connection to own DB

Pkrasota:


ViktoriiaSemenchenko (changes requested):
incredible work. One note - make a separate dao/impl package for ManufacturerDaoImpl

ViktoriiaSemenchenko (approved):
good luck)

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,120 @@
...
+            deleteStatement.setLong(1, id);
+            return deleteStatement.executeUpdate() > 0;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't delete manufacturer from DB with id: "
+                    + id, e);
+        }
+    }
+
+    private Manufacturer getManufacturer(ResultSet resultSet) throws SQLException {
+        Manufacturer manufacturer = new Manufacturer();
```
maxline, line 114: You can use constructor with parameters, instead of default constructor with setter.

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,23 @@
...
-    public static void main(String[] args) {
 
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
+    public static void main(String[] args) {
+        ManufacturerDao manufacturerDao =
+                (ManufacturerDao) injector.getInstance(ManufacturerDao.class);
+        Manufacturer manufacturer = new Manufacturer();
+        manufacturer.setName("Mercedes-Benz");
+        manufacturer.setCountry("Germany");
```
maxline, line 16: You can use constructor with parameters, instead of default constructor with setter.

maxline (approved):
Well done! Left two minor comments. Please take a look for the future.

PR #649, opened 2023-06-09
Implemented task jdbc

Valdemar619:


vladyslavkryvets (approved):
Good job!

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,113 @@
...
+                PreparedStatement preparedStatement = connection.prepareStatement(GET_BY_ID)) {
+            preparedStatement.setLong(1, id);
+            ResultSet resultSet = preparedStatement.executeQuery();
+            Optional<Manufacturer> result;
+            if (resultSet.next()) {
+                result = Optional.of(createManufacturer(resultSet));
+            } else {
+                result = Optional.empty();
+            }
+            return result;
```
maxline, line 60: refactor in this way, and simplify `if` construction above
```suggestion
            return Optional.ofNullable(result);
```

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,113 @@
...
+                PreparedStatement preparedStatement = connection.prepareStatement(DELETE_BY_ID)) {
+            preparedStatement.setLong(1, id);
+            return preparedStatement.executeUpdate() > 0;
+        } catch (Exception e) {
+            throw new DataProcessingException("Can't delete manufacturer by id " + id, e);
+        }
+    }
+
+    private Manufacturer createManufacturer(ResultSet resultSet) throws SQLException {
+        Manufacturer manufacturer = new Manufacturer("name", "country");
```
maxline, line 109: This is not correct piece of code. You should get `name` and `country` values from `resultSet`.
Also you can add a constructor with three parameters and use it here, instead of using setter for setting `id` value.

maxline (changes requested):
Good job! let's improve the solution

sarakhmen (approved):
Good job :)

PR #650, opened 2023-06-09
create jdbc

DV7Art:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,29 @@
...
+        manufacturerDao.create(volvo);
+        manufacturerDao.create(bmw);
+        manufacturerDao.getAll()
+                .stream()
+                .forEach(System.out::println);
+        manufacturerDao.delete(bmw.getId());
+        manufacturerDao.getAll()
+                .stream()
+                .forEach(System.out::println);
     }
```
AnastasiiaVariuchuk, line 28: -

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,29 @@
...
+        Manufacturer hyundai = new Manufacturer("Hyundai", "Korea");
+        Manufacturer volvo = new Manufacturer("Volvo", "Sweden");
+        Manufacturer bmw = new Manufacturer("BMW", "Germany");
+        manufacturerDao.create(zaz);
+        manufacturerDao.create(hyundai);
+        manufacturerDao.create(volvo);
+        manufacturerDao.create(bmw);
+        manufacturerDao.getAll()
+                .stream()
+                .forEach(System.out::println);
```
maxline, line 23: This code can be simplified
```suggestion
        manufacturerDao.getAll().forEach(System.out::println);
```
Fix all cases please.

maxline (approved):
Well done! Approved with a small suggestion. Please take a look for the future.

src/main/java/mate/jdbc/Main.java
DV7Art: I thought it was more readable.

PR #661, opened 2023-06-11
main() method works correctly, you may not believe but looks like eve…

MykNich:
…rything ok wow :o

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -1,7 +1,26 @@
 package mate.jdbc;
 
+import mate.jdbc.dao.ManufacturerDao;
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Manufacturer;
+
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
+        ManufacturerDao manufacturerDao = (ManufacturerDao)
+                injector.getInstance(ManufacturerDao.class);
+        Manufacturer manufacturer = new Manufacturer("Chevrolet", "USA");
+        System.out.println(manufacturerDao.getAll());
+        System.out.println(manufacturerDao.create(manufacturer));
+
+        manufacturer.setCountry("DAEWOO-ZAZ");
+        manufacturer.setName("Ukraine");
+        System.out.println(manufacturerDao.update(manufacturer));
+        System.out.println(manufacturerDao.delete(6L));
+
+        System.out.println(manufacturerDao.get(1L));
+        System.out.println(manufacturerDao.getAll());
 
     }
 }
--- a/src/main/java/mate/jdbc/dao/ManufacturerDao.java
+++ b/src/main/java/mate/jdbc/dao/ManufacturerDao.java
@@ -1,0 +1,18 @@
+package mate.jdbc.dao;
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.model.Manufacturer;
+
+public interface ManufacturerDao {
+
+    Manufacturer create(Manufacturer manufacturer);
+
+    Optional<Manufacturer> get(Long id);
+
+    List<Manufacturer> getAll();
+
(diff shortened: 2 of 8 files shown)


mykyta-udav (approved):
Good job

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,26 @@
...
+        Manufacturer manufacturer = new Manufacturer("Chevrolet", "USA");
+        System.out.println(manufacturerDao.getAll());
+        System.out.println(manufacturerDao.create(manufacturer));
+
+        manufacturer.setCountry("DAEWOO-ZAZ");
+        manufacturer.setName("Ukraine");
+        System.out.println(manufacturerDao.update(manufacturer));
+        System.out.println(manufacturerDao.delete(6L));
+
+        System.out.println(manufacturerDao.get(1L));
```
sarakhmen, line 22: You are already aware of these issues :)

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,26 @@
...
+        System.out.println(manufacturerDao.getAll());
```
sarakhmen, line 24:
```suggestion
```

src/main/java/mate/jdbc/dao/ManufacturerDao.java
```diff
@@ -0,0 +1,18 @@
+package mate.jdbc.dao;
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.model.Manufacturer;
+
+public interface ManufacturerDao {
+
```
sarakhmen, line 8:
```suggestion
```

src/main/java/mate/jdbc/dao/ManufacturerDaoImp.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.exception.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class ManufacturerDaoImp implements ManufacturerDao {
+    private static final String GET_ALL
```
sarakhmen, line 17: Let's add QUERY suffix to your constants here so that their purpose is more clear

src/main/java/mate/jdbc/dao/ManufacturerDaoImp.java
```diff
@@ -0,0 +1,108 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement statement = connection.prepareStatement(
+                        CREATE, PreparedStatement.RETURN_GENERATED_KEYS)) {
+            statement.setString(1, manufacturer.getName());
+            statement.setString(2, manufacturer.getCountry());
+            statement.executeUpdate();
+            ResultSet keys = statement.getGeneratedKeys();
+            if (keys.next()) {
+                Long id = keys.getObject(1, Long.class);
+                manufacturer.setId(id);
```
sarakhmen, line 39: let's inline it

src/main/java/mate/jdbc/dao/ManufacturerDaoImp.java
```diff
@@ -0,0 +1,108 @@
...
+        } catch (Exception e) {
+            throw new DataProcessingException("Can't delete manufacturer by id " + id, e);
+        }
+    }
+
+    private Manufacturer createManufacturer(ResultSet resultSet) throws SQLException {
+        Manufacturer manufacturer = new Manufacturer(
+                resultSet.getString("name"),
+                resultSet.getString("country"));
+        manufacturer.setId(resultSet.getObject("id", Long.class));
```
sarakhmen, line 105: As described in common mistakes: When you convert ResultSet to Manufacturer better create an object using setters or constructor but not both of them, because it’s not consistent to use both ways of initialization of object.

sarakhmen (changes requested):
Let's improve your solution :)

PR #675, opened 2023-06-26
Implemented JBDC

vaniii1:


src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,118 @@
...
+                String name = resultSet.getString("name");
+                String country = resultSet.getString("country");
+                manufacturer.setId(id);
+                manufacturer.setName(name);
+                manufacturer.setCountry(country);
+                return Optional.of(manufacturer);
+            }
+            return Optional.empty();
+        } catch (SQLException e) {
+            throw new RuntimeException("Can't get manufacturer by id from DB", e);
```
olekskov, line 60: Replace RuntimeException with a custom exception called DataProcessingException as mentioned in the task description.

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,118 @@
...
+                        connection.prepareStatement(getManufacturerRequest)) {
+            preparedStatement.setLong(1, id);
+            Manufacturer manufacturer = new Manufacturer();
+            ResultSet resultSet = preparedStatement.executeQuery();
+            if (resultSet.next()) {
+                String name = resultSet.getString("name");
+                String country = resultSet.getString("country");
+                manufacturer.setId(id);
+                manufacturer.setName(name);
+                manufacturer.setCountry(country);
```
olekskov, line 55: This logic is repeated for get() and getAll() methods.
Please extract to separate private method and use it in both public methods

Ivan95kos (approved):
Nice

PR #687, opened 2023-07-16
ManufacturerDao is created

YaroslavKostenko89:


src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,120 @@
...
+import mate.jdbc.dao.exception.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class ManufacturerDaoImpl implements ManufacturerDao {
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        String insertManufacturerRequest = "INSERT INTO manufacturer(name, country) values(?, ?);";
```
Ivan95kos, line 20: It is better to write all SQL commands in upper case
```suggestion
        String insertManufacturerRequest = "INSERT INTO manufacturer(name, country) VALUES(?, ?);";
```

Ivan95kos (approved):
Great job, with minor comment

PR #690, opened 2023-07-19
Hw jdbc

AntonU91:


src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,137 @@
...
+                        connection.prepareStatement(deleteQuery)) {
+            deleteStatement.setLong(1, id);
+            changedRowCount = deleteStatement.executeUpdate();
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can not delete manufacturer by id " + id, e);
+        }
+        return changedRowCount > 0;
+    }
+
+    private void enrichManufacturerObj(Manufacturer manufacturer, ResultSet resultSet)
```
okuzan, line 116: enrich isn't a good option here. let's call it parseManufacturer or getManufacturer or parseManufacturerFromResultSet etc

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,137 @@
...
+    @Override
+    public List<Manufacturer> getAll() {
+        List<Manufacturer> manufacturerList = new ArrayList<>();
+        String selectAllQuery = "SELECT *"
+                + " FROM manufacturers"
+                + " WHERE is_deleted='FALSE'";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getAllManufacturesStatement =
+                        connection.prepareStatement(selectAllQuery)) {
+            getAllManufacturesStatement.execute();
```
okuzan, line 69: don't use execute in this task, use executeQuery

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,137 @@
...
+        }
+        return Optional.empty();
+    }
+
+    @Override
+    public List<Manufacturer> getAll() {
+        List<Manufacturer> manufacturerList = new ArrayList<>();
+        String selectAllQuery = "SELECT *"
+                + " FROM manufacturers"
+                + " WHERE is_deleted='FALSE'";
```
okuzan, line 65:
```suggestion
                + " WHERE is_deleted=FALSE";
```
we don't need to escape false

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,137 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getAllManufacturesStatement =
+                        connection.prepareStatement(selectAllQuery)) {
+            getAllManufacturesStatement.execute();
+            ResultSet resultSet = getAllManufacturesStatement.getResultSet();
+            while (resultSet.next()) {
+                Manufacturer manufacturer = new Manufacturer();
+                enrichManufacturerObj(manufacturer, resultSet);
```
okuzan, line 73: instead of passing an empty object, remove this from arguments of the method, and create it inside of the method

okuzan (changes requested):

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
AntonU91: Done

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
AntonU91: Done

src/main/resources/init_db.sql
```diff
@@ -0,0 +1,10 @@
+CREATE DATABASE `taxi_service_db` /*!40100 DEFAULT CHARACTER SET utf8mb3 */ /*!80016 DEFAULT ENCRYPTION='N' */;
+
+CREATE TABLE `manufacturers` (
+                                 `id` bigint NOT NULL AUTO_INCREMENT,
+                                 `name` varchar(45) DEFAULT NULL,
+                                 `country` varchar(45) DEFAULT NULL,
+                                 `is_deleted` varchar(5) DEFAULT 'FALSE',
```
Ivan95kos, line 7:
```suggestion
                                 `is_deleted` tinyint NOT NULL DEFAULT '0',
```

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
Ivan95kos: This is because you used varchar and you need tinyint, please fix it

Ivan95kos (changes requested):
Good job overall!

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,133 @@
...
+            throw new DataProcessingException("Can not update manufacturer " + manufacturer, e);
+        }
+        return manufacturer;
+    }
+
+    @Override
+    public boolean delete(Long id) {
+        int changedRowCount;
+        String deleteQuery = "UPDATE manufacturers "
+                + "SET is_deleted='1'"
```
kozhukhovsky, line 97:
```suggestion
                + "SET is_deleted=TRUE"
```

kozhukhovsky (approved):

PR #693, opened 2023-07-20
Hw jv jdbc

BakhmetIvan:


src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.exceptions.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class ManufacturerDaoImpl implements ManufacturerDao {
+    private static final int FIRST_VALUE = 1;
```
kozhukhovsky, line 18: this is probably the case when I would not take out indexes in contants
it does not make the text more informative

kozhukhovsky (approved):

kozhukhovsky (approved):

PR #694, opened 2023-07-21
jv-jdbc hometask completed

MarynaKushniruk:


src/main/java/mate/jdbc/dao/ManufacturerDaoImp.java
```diff
@@ -0,0 +1,121 @@
...
+            ResultSet resultSet = statement.executeQuery();
+            if (resultSet.next()) {
+                return Optional.of(getManufacturer(resultSet));
+            } else {
+                return Optional.empty();
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get Manufacturer with ID" + id
+                    + " from DB", e);
+        }
```
okuzan, line 108:
```suggestion
            if (resultSet.next()) {
                return Optional.of(getManufacturer(resultSet));
            }
        } catch (SQLException e) {
            throw new DataProcessingException("Can't get Manufacturer with ID" + id
                    + " from DB", e);
        }
        return Optional.empty();
```

src/main/java/mate/jdbc/dao/ManufacturerDaoImp.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    private static Manufacturer getManufacturer(ResultSet resultSet) throws SQLException {
+        Manufacturer manufacturer = new Manufacturer();
+        String name = resultSet.getString("name");
+        String country = resultSet.getString("country");
+        Long id = resultSet.getObject("id", Long.class);
+        manufacturer.setId(id);
+        manufacturer.setName(name);
+        manufacturer.setCountry(country);
```
okuzan, line 118: use retrieved values immediately, without saving to temp local vars

okuzan (changes requested):
Good job!

Ivan95kos (approved):
Nice

PR #701, opened 2023-07-31
Solution

Volodymyr-Mykychak:


src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,119 @@
...
+        String sql = "DELETE FROM manufacturers WHERE id = ?";
+        try (Connection conn = ConnectionUtil.getConnection();
+                PreparedStatement stmt = conn.prepareStatement(sql)) {
+            stmt.setLong(1, id);
+            int deletedRows = stmt.executeUpdate();
+            return deletedRows > 0;
+        } catch (SQLException ex) {
+            throw new DataProcessingException("Can't delete manufacturer by id from DB "
+                    + id, ex);
+        }
```
okuzan, line 117: You should implement delete using soft delete technique, as discussed in task description. That means here you will update is_deleted column, which you have in init_db but never use

okuzan (changes requested):

src/main/java/mate/jdbc/dao/ManufacturerDaoImpl.java
```diff
@@ -0,0 +1,119 @@
...
+        } catch (SQLException ex) {
+            throw new DataProcessingException("Can't insert manufacturer to DB "
+                    + manufacturer, ex);
+        }
+        return manufacturer;
+    }
+
+    @Override
+    public Optional<Manufacturer> get(Long id) {
+        String sql = "SELECT * FROM manufacturers WHERE id = ?";
```
olekskov, line 44: get and update operations shouldn't work with deleted entities (like done for getAll)

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,52 @@
...
+    @Override
+    public Manufacturer createManufacturer(Manufacturer manufacturer) {
+        // Перевірка валідності даних виробника перед додаванням до бази даних
+        validationService.validateBeforeCreate(manufacturer);
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer getManufacturerById(Long id) {
+        return manufacturerDao.get(id).orElse(null);
```
olekskov, line 34: Returning null is generally not a good practice
Please use .orElseThrow() method with proper exception instead

olekskov (changes requested):

olekskov (approved):
Good job

